PR #32, opened 2024-11-20
I2C Driver Implementation with RTOS STM32 HAL

taisirhassan:
**I2C Module Implementation**
Implements RTOS-friendly I2C interface for processor board sensors. Key features:

Interrupt-based transfers (no polling)
Configurable timeouts per device
Simple peripheral-based interface (no need to manage HAL handles)
One device per bus design from latest board revision


Redstone-ray:
I know this is still WIP, but IMU, BARO and MAG are not three different peripherals, please double check the relevant documentation to understand which three IMUs we are using

taisirhassan:
> I know this is still WIP, but IMU, BARO and MAG are not three different peripherals, please double check the relevant documentation to understand which three IMUs we are using

Updated to use the right IMUs, had a general one that I forgot to change. Thanks for pointing it out.

Redstone-ray (commented):
Thanks for your update, quick question

- you indicated a max len of "max 255 byte", but len is a uint16_t, which can exceed your limit of 255?
- should the data be modifiable? if not, declare it as a constant

taisirhassan:
> Thanks for your update, quick question
> 
> * you indicated a max len of "max 255 byte", but len is a uint16_t, which can exceed your limit of 255?
> * should the data be modifiable? if not, declare it as a constant

You're right on both points. I changed len to uint8_t to match the 255-byte limit and made data const in write_reg() since it's read-only. I kept data non-const in read_reg() since it needs to be written to the buffer. Thanks.

src/drivers/i2c/i2c.h
```diff
@@ -0,0 +1,22 @@
...
+    I2C_DEVICE_COUNT // Number of devices
+} i2c_device_t;
+
+// Must be called before RTOS scheduler starts
+w_status_t i2c_init(i2c_device_t device, uint32_t timeout_ms);
+
+// Read len bytes from reg into data buffer, max 255 byte
+w_status_t i2c_read_reg(i2c_device_t device, uint8_t reg, uint8_t *data, uint8_t len);
+
+w_status_t i2c_write_reg(i2c_device_t device, uint8_t reg, const uint8_t *data, uint8_t len);
```
celery6, line 20: add comments abt this function. sure it might seem obvious by inspection but imo all public interfaces should be over-described rather than omitting even the obvious.

src/drivers/i2c/i2c.h
```diff
@@ -0,0 +1,22 @@
...
+
+typedef enum {
+    I2C_DEVICE_POLULU, // Polulu AltIMU-v6
+    I2C_DEVICE_MOVELLA, // Movella mti-630 IMU
+    I2C_DEVICE_ST, // ST LSM6DSO32 IMU
+    I2C_DEVICE_COUNT // Number of devices
+} i2c_device_t;
+
+// Must be called before RTOS scheduler starts
+w_status_t i2c_init(i2c_device_t device, uint32_t timeout_ms);
```
celery6, line 15: comment more. dont expect users to reference the design doc when using this function, it should be explained fully in code. (what it does, what each param is etc)

celery6 (changes requested):
few nits. also left a comment in design doc abt updating the diagram. otherwise this looks great

src/drivers/i2c/i2c.h
```diff
@@ -0,0 +1,37 @@
...
+// Supported I2C devices
+typedef enum {
+    I2C_DEVICE_POLULU, // Polulu AltIMU-v6
+    I2C_DEVICE_MOVELLA, // Movella MTi-630 IMU
+    I2C_DEVICE_ST, // ST LSM6DSO32 IMU
+    I2C_DEVICE_COUNT // Number of devices
+} i2c_device_t;
+
+// Initialize an I2C device
+// device: I2C device to initialize
```
Joe-Joe-Joe-Joe, line 16: If we're commenting in this level of detail might as well use doxygen

Joe-Joe-Joe-Joe:
How exactly do you register the callbacks? Are you going to write a dedicated callback for each i2c device, then have the init function pass those function pointers behind the scenes? Would it be cleaner (since you callbacks are just one-liners to give the correct semaphore aiui) to implement the global callback function and a simple switch-case based on which handle called the callback?

src/drivers/i2c/i2c.h
```diff
@@ -0,0 +1,37 @@
+#ifndef I2C_H
+#define I2C_H
+
+#include "common.h" // Common definitions
+#include <stdint.h> // Standard integer types
+
+// Supported I2C devices
+typedef enum {
+    I2C_DEVICE_POLULU, // Polulu AltIMU-v6
+    I2C_DEVICE_MOVELLA, // Movella MTi-630 IMU
```
Joe-Joe-Joe-Joe, line 10: the movella talks over UART, not I2C. Any references to it in I2C should be removed

taisirhassan:
> How exactly do you register the callbacks? Are you going to write a dedicated callback for each i2c device, then have the init function pass those function pointers behind the scenes? Would it be cleaner (since you callbacks are just one-liners to give the correct semaphore aiui) to implement the global callback function and a simple switch-case based on which handle called the callback?

I was planning on using a single global callback for the i2c module, as you said it's just the same callback logic just on different semaphores it would be a lot less repetitive.

src/drivers/i2c/i2c.h
taisirhassan: honestly had no idea on what level of detail I should've utilized but would doxygen be good atp if we're assuming the user isn't familiar with the design document? @celery6

celery6: any code on its own should always contain documentation regardless of whether design doc exists or not. 

Idc about doxygen or any other style tbh but sure lets make doxygen style our standard from now on

celery6 (approved):
could change comment style to doxygen otherwise LGTM!!!

PR #36, opened 2024-11-26
ADC module header file

jerryjcr:
Header file for the ADC module

- defines enum for different adc channels
- defines public interfaces


Joe-Joe-Joe-Joe:
Comments in the gdoc

src/drivers/adc/adc.h
```diff
@@ -0,0 +1,25 @@
...
+
+#include <stdbool.h>
+
+typedef enum {
+    PROCESSOR_BOARD_VOLTAGE = 0,
+    ADC_CHANNEL_2,
+    ADC_CHANNEL_3,
+    ADC_CHANNEL_4,
+    ADC_CHANNEL_5
+} adc_channel_t;
```
celery6, line 12: we should hide these unused channel from public view since it gives a false promise that they'd actually work.

src/drivers/adc/adc.h
```diff
@@ -0,0 +1,25 @@
...
+
+w_status_t adc_init(void);
+
+// writes adc value of the specified channel to the output parameter
+w_status_t adc_get_value(adc_channel_t channel, uint32_t *output);
+
+// writes processor board voltage to the output parameter
+w_status_t adc_get_board_voltage(uint16_t *output);
+
+// writes processor board voltage to the output parameter
```
celery6, line 22: comment shld mention blocking behavior - if it blocks or not, for how long, is it thread-safe (can multiple threads use it at once without dying)

celery6, line 22: specify units !!

celery6 (changes requested):
also left some comments in google doc

Joe-Joe-Joe-Joe (dismissed):
Make sure health checks documentation is clear that it is now responsible for getting ADC counts and converting them

Joe-Joe-Joe-Joe:
Also, as mentioned in the doc, please document the ADC configuration settings used in the .ioc file. There are many of them.

src/drivers/adc/adc.h
```diff
@@ -0,0 +1,17 @@
+#ifndef ADC_H
+#define ADC_H
+
+#include <stdbool.h>
+
```
celery6, line 5: nit: add `#include "rocketlib/includes/common.h"` since u used w_status_t

idk why omitting it didnt throw a compile warning.. i will look into that.

src/drivers/adc/adc.h
```diff
@@ -0,0 +1,17 @@
+#ifndef ADC_H
+#define ADC_H
+
+#include <stdbool.h>
+
+typedef enum {
+    PROCESSOR_BOARD_VOLTAGE = 0,
+} adc_channel_t;
```
celery6, line 8: nit: imo any enum that might have more values added to it later should have an extra enum for the max value. example:
image

celery6 (dismissed):
2 minor requests otherwise lgtm!!

celery6 (approved):

Issue #4, opened 2024-10-13
disable config_enable_backwards_compatibility


celery6:
We noticed last year's codebase uses a few legacy types from FreeRTOS v8, like xQueueHandle instead of the new v10 names. See https://ww3.freertos.org/Documentation/04-Roadmap-and-release-note/02-Release-notes/01-FreeRTOS-V8#new-freertos-defined-typedefs-names

Check that we have no reason to use old type names. If confirmed, disable this config thing to force everyone to use the latest type names.

PR #39, opened 2024-12-17
add gpio header file

celery6:
Design details are in the design doc under the gpio module section


src/drivers/gpio/gpio.h
```diff
@@ -0,0 +1,46 @@
...
+    GPIO_PIN_GREEN_LED,
+    GPIO_PIN_BLUE_LED,
+    GPIO_PIN_COUNT // Enum max value
+} gpio_pin_t;
+
+/**
+ * Read the current level of `pin` into `level`. Non-blocking. Not thread-safe
+ * if multiple threads interact with the same pin.
+ */
+w_status_t gpio_read(gpio_pin_t pin, gpio_level_t *level);
```
Armaan-Sengupta, line 32: Does the hal have a fail condition here? if not for simplicity it might be simpler to just return the value. If that is the case pls put the output first

Armaan-Sengupta:
Regarding my inline comment, the hal read just returns the value. So is the idea here going to be to check the other statuses and initiations of the pin to make sure it is even valid to read from in the first place, and if not then return a fail, because the hal gpio read command itself won't fail, or at least we won't know it failed.

Will the port/pin map to the GPIOs be two other structs that are index aligned to eachother and the actual GPIOs?

src/drivers/gpio/gpio.h
celery6: hal doesnt have a fail but its possible that the `pin` input is invalid

celery6: or if the cmd required any further processing in the future (for example we decide to add mutex), it would have even more fail conditions. Then we'd have to go back and change this to return status and everything using it.

Yeah that's unlikely but returning status has no downsides only future-proofing benefits imo

celery6:
> Regarding my inline comment, the hal read just returns the value. So is the idea here going to be to check the other statuses and initiations of the pin to make sure it is even valid to read from in the first place, and if not then return a fail, because the hal gpio read command itself won't fail, or at least we won't know it failed.
> 
> Will the port/pin map to the GPIOs be two other structs that are index aligned to eachother and the actual GPIOs?

from design doc just added: `Module will store an array mapping gpio_pin enum indexes to each one’s stm32 hal port+pin.`

taisirhassan:
Just curious but would we want to consider adding mutexes later on for future-proof purposes? If the STM32 HAL functions are atomic there would be no need correct? We could also use validation checks or higher-level synchronization outside the GPIO module to ensure safe access.

celery6:
> Just curious but would we want to consider adding mutexes later on for future-proof purposes? If the STM32 HAL functions are atomic there would be no need correct? We could also use validation checks or higher-level synchronization outside the GPIO module to ensure safe access.

hal gpio is atomic but wrapping it makes these functions not atomic so technically they should have mutexes. I chose to skip that because seems overkill... but idk maybe we should just add it now?

celery6:
added change to also address #4

Armaan-Sengupta (approved):
LGTM

Joe-Joe-Joe-Joe (approved):
Header looks good. I don't know what all the other changed files are about, for the sake of purity shouldn't those changes be handled in another PR? Idc regardless, just curious.

celery6:
> Header looks good. I don't know what all the other changed files are about, for the sake of purity shouldn't those changes be handled in another PR? Idc regardless, just curious.

youre right they should be. I got lazy and put the 1-line changes in this pr since its already a really light pr. They are in separate commits at least which is more important imo

PR #42, opened 2024-12-19
Add AltIMU10 header file

arpitarunkumaar:
Design details are laid out in the module design doc under the AltIMU section [here](https://docs.google.com/document/d/1OokvdQ8c74xzt_znfmBUrUzJ5vtqvndMz0l9xxN98dY/edit?tab=t.0#heading=h.mxwft7budhpj). 

**Comments:**
- We understand your point about keeping register addresses local to AltIMU_Init() since they are only needed within that function. Making them global is unnecessary as AltIMU_Init() is called by the reset function anyway. We believe this local approach is efficient and avoids unnecessary and excessive clutter in the global defines section. Let us know if you feel otherwise or have any concerns.
- Do we need the externs to the protocol handlers?

### DIFF
--- a/src/drivers/altimu-10/altimu-10.h
+++ b/src/drivers/altimu-10/altimu-10.h
@@ -1,55 +1,49 @@
 #ifndef ALTIMU_10_H
 #define ALTIMU_10_H
 
-#include "rocketlib/include/common.h"       
+#include "rocketlib/include/common.h"
 #include <stdbool.h>
 #include <stdint.h>
 
 // Sensor data structure definitions
-typedef struct
-{
-     int16_t x;
-     int16_t y;
-     int16_t z;
+typedef struct {
+  int16_t x;
+  int16_t y;
+  int16_t z;
 } altimu_accelerometer_data;
 
-typedef struct
-{
-     int16_t x;
-     int16_t y;
-     int16_t z;
+typedef struct {
+  int16_t x;
+  int16_t y;
+  int16_t z;
 } altimu_gyroscope_data;
 
-typedef struct
-{
-     int16_t x;
-     int16_t y;
-     int16_t z;
+typedef struct {
+  int16_t x;
+  int16_t y;
+  int16_t z;
 } altimu_magnetometer_data;
 
-typedef struct
-{
-     float pressure;
-     float temperature;
+typedef struct {
+  float pressure;
+  float temperature;
 } altimu_barometer_data;
 
 // IMU device addresses and configuration
-typedef struct
-{
-     uint8_t acc_addr;
-     uint8_t gyr_addr;
-     uint8_t mag_addr;
-     uint8_t bar_addr;
-     uint8_t who_am_i_reg;
+typedef struct {
+  uint8_t acc_addr;
+  uint8_t gyr_addr;
+  uint8_t mag_addr;
+  uint8_t bar_addr;
+  uint8_t who_am_i_reg;
 } altimu_imu_address_info;
 
 // Define the communication protocols
-typedef enum
-{
-     ALTIMU_I2C_PROTOCOL,
-     ALTIMU_SPI_PROTOCOL,
-     ALTIMU_UART_PROTOCOL,
-     ALTIMU_UNKNOWN_PROTOCOL
+typedef enum {
+  ALTIMU_I2C_PROTOCOL,
+  ALTIMU_SPI_PROTOCOL,
+  ALTIMU_UART_PROTOCOL,
+  ALTIMU_UNKNOWN_PROTOCOL
 } altimu_protocol;
 
 // Function prototypes
@@ -63,10 +57,12 @@ w_status_t altimu_init(I2C_HandleTypeDef *hi2c);
 
 /**
  * @brief Detects the communication protocol and IMU addresses.
- * @param imu_info Pointer to altimu_imu_address_info structure to store detected addresses.
+ * @param imu_info Pointer to altimu_imu_address_info structure to store
+ * detected addresses.
  * @return Detected protocol type.
  */
-altimu_protocol altimu_detect_protocol_and_addresses(altimu_imu_address_info *imu_info);
+altimu_protocol
+altimu_detect_protocol_and_addresses(altimu_imu_address_info *imu_info);
 
 /**
  * @brief Retrieves accelerometer data.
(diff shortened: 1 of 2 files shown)


src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+} altimu_barometer_data;
+
+// AltIMU device addresses and configuration
+typedef struct {
+    uint8_t acc_addr;
+    uint8_t gyr_addr;
+    uint8_t mag_addr;
+    uint8_t bar_addr;
+    uint8_t who_am_i_reg;
+} altimu_imu_address_info;
```
celery6, line 39: does this data need to be revealed in this header file? what else uses it

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+
+// Function prototypes
+
+/**
+ * @brief Initializes the Pololu AltIMU module.
+ * @param hi2c Pointer to the I2C handle.
+ * @return Status of the operation.
+ */
+w_status_t altimu_init(I2C_HandleTypeDef *hi2c);
```
celery6, line 48: i2c is taken care of by the i2c module - see comments in gdoc

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+ * @param data Pointer to the data to write.
+ * @return Status of the operation.
+ */
+w_status_t altimu_write_register(uint8_t reg_addr, const void *data);
+
+/**
+ * @brief Delays execution for a specified number of milliseconds.
+ * @param ms Number of milliseconds to delay.
+ */
+void altimu_delay_ms(uint32_t ms);
```
celery6, line 135: should these be revealed in this header file? seems like only the altimu module will use this?

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+
+// Constants and macros definitions
+#define ALTIMU_WHO_AM_I_REG 0x0F // WHO_AM_I register address
+#define ALTIMU_EXPECTED_WHO_AM_I 0x6C // Expected WHO_AM_I value
+
+// Below are subject to change...
+extern I2C_HandleTypeDef hi2c1; // I2C handler
+extern SPI_HandleTypeDef hspi1; // SPI handler
+extern UART_HandleTypeDef huart2; // UART handler
```
celery6, line 144: wont need these since these are taken care of by the respective device driver modules for i2c, uart (we dont have anything on spi yet)

celery6 (changes requested):
also added some comments in gdoc.

i like how clean this is - gg for great clarity

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+    int16_t x;
+    int16_t y;
+    int16_t z;
+} altimu_gyroscope_data;
+
+typedef struct {
+    int16_t x;
+    int16_t y;
+    int16_t z;
+} altimu_magnetometer_data;
```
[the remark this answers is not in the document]
Joe-Joe-Joe-Joe: I don't think so. Here, the data is going to come out in a fraction of full scale. The particular IMU module should then math that into a float by applying the correct conversion factor (data / 2^16 * 1000 dps) If your full scale is +/- 1000 dps.  By doing this we ensure all sensor data is provided in a common format at the source. 

I think every "vector" that we pass around should be floating point type by convention, and that will prevent accidental integer math errors.

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+ * @param hi2c Pointer to the I2C handle.
+ * @return Status of the operation.
+ */
+w_status_t altimu_init(I2C_HandleTypeDef *hi2c);
+
+/**
+ * @brief Retrieves accelerometer data.
+ * @return Accelerometer data as altimu_accelerometer_data structure.
+ */
+altimu_accelerometer_data altimu_get_acc_data(void);
```
Joe-Joe-Joe-Joe, line 54: See my comments in the doc, I think these should be returning status codes and modifying pointers rather than returning structs

src/drivers/altimu-10/altimu-10.h
Joe-Joe-Joe-Joe, line 135: I don't see this function in the gdoc, what is its purpose? I don't know what convention has been adopted but our delay function should be common (such as freeRTOS `vTaskDelay`) across modules

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * @brief Delays execution for a specified number of milliseconds.
+ * @param ms Number of milliseconds to delay.
+ */
+void altimu_delay_ms(uint32_t ms);
+
+// Constants and macros definitions
+#define ALTIMU_WHO_AM_I_REG 0x0F // WHO_AM_I register address
+#define ALTIMU_EXPECTED_WHO_AM_I 0x6C // Expected WHO_AM_I value
```
Joe-Joe-Joe-Joe, line 139: Why do these need to be exposed in the header? Would one not simply call `altimu_check_sanity` and have the register comparisons happen under the hood?

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,146 @@
...
+typedef struct {
+    int16_t x;
+    int16_t y;
+    int16_t z;
+} altimu_magnetometer_data;
+
+typedef struct {
+    float pressure;
+    float temperature;
+} altimu_barometer_data;
```
Joe-Joe-Joe-Joe, line 30: Do the device registers for pressure and temp store data as a float? Otherwise I assume these would be ints like the IMU data?

Joe-Joe-Joe-Joe (changes requested):

src/drivers/altimu-10/altimu-10.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <stdbool.h>
+#include <stdint.h>
+
+#include "common/math/math.h"
+#include "rocketlib/include/common.h"
+
+typedef struct {
+    float pressure; // pascal
+    float temperature; // celsius
```
FinnBreu, line 12: We discussed a few weeks ago that the temperature measurement would not be reliable: the sensor is on a board inside the electronics bay, which heats up more than air temp outside the rocket

celery6: this module is a device driver. its purpose is to allow other ppl to collect all the data possible. IMU Handler will choose whether to ignore temperature

FinnBreu (approved):

PR #43, opened 2024-12-20
Unit Test Build System With GTest & FFF

taisirhassan:


Joe-Joe-Joe-Joe:
If/when the build script is ready, please add VERY explicit instructions on
- Setting up build environment and toolchain
- Using the build/run script

Joe-Joe-Joe-Joe:
On the CI/CD, can we add the job to clean/build/test on this PR as well? if we do, it would be nice to export the results. At work we use JUnit for this. Once you generate the test report you can use smth like this https://github.com/marketplace/actions/junit-report-action to display the results, seemingly inline? That can probably wait for another PR but the workflow to make sure tests pass on PRs would good to have now.

taisirhassan:
> On the CI/CD side, can we add the job to clean/build/test on this PR as well? if we do, it would be nice to export the results. At work we use JUnit for this. Once you generate the test report (e.g. `ctest --output-junit testRes.xml`) you can use smth like this https://github.com/marketplace/actions/junit-report-action to display the results, seemingly inline? That can probably wait for another PR but the workflow to make sure tests pass on PRs would good to have now.

Yeah, I was thinking of adding a job for building and testing later. Will look into generating a test report in another PR, I think making sure the workflow works first would be good for now.

Joe-Joe-Joe-Joe:
Some notes about the Windows install
- Using Git bash (most people's default) required running it as admin?
- I also had to get rid of `-G "MinGW Makefiles"`, even though I was definitely using the mingw toolchain
- The clang formatter requires you to have LLVM installed, so everyone is going to have that as well

cmake/stm32cubemx/CMakeLists.txt
celery6, line 1: this file was autogen by cubemx except where i commented

CMakeLists.txt
celery6, line 1: this was autogen by cubemx but modified a lot manually

CMakePresets.json
celery6, line 1: rn these presets arent incredibly useful since our build configs are similar and gtest doesnt use this.. is it better to delete this file and manually configure cmake settings in run.sh?

scripts/run.sh
celery6, line 1: i mashed gtest and binary building into this one script based on @taisirhassan test.sh.
Alternatively we could have a test.sh and a build.sh, but theres a lot of overlap.
Is this more or less confusing? i think the problem with having all in 1 is it implies testing and building binary are the same build config, but they are entirely separate. Maybe thats a good thing to hide in a script though? no one really needs to know that in practice

processor_board_Debug.launch
```diff
@@ -5,14 +5,14 @@
...
     <booleanAttribute key="com.st.stm32cube.ide.mcu.debug.launch.enable_swv" value="false"/>
     <intAttribute key="com.st.stm32cube.ide.mcu.debug.launch.formatVersion" value="2"/>
     <stringAttribute key="com.st.stm32cube.ide.mcu.debug.launch.ip_address_local" value="localhost"/>
     <booleanAttribute key="com.st.stm32cube.ide.mcu.debug.launch.limit_swo_clock.enabled" value="false"/>
     <stringAttribute key="com.st.stm32cube.ide.mcu.debug.launch.limit_swo_clock.value" value=""/>
-    <stringAttribute key="com.st.stm32cube.ide.mcu.debug.launch.loadList" value="{&quot;fItems&quot;:[{&quot;fIsFromMainTab&quot;:true,&quot;fPath&quot;:&quot;Debug/processor_board.elf&quot;,&quot;fProjectName&quot;:&quot;processor_board&quot;,&quot;fPerformBuild&quot;:true,&quot;fDownload&quot;:true,&quot;fLoadSymbols&quot;:true}]}"/>
+    <stringAttribute key="com.st.stm32cube.ide.mcu.debug.launch.loadList" value="{&quot;fItems&quot;:[{&quot;fIsFromMainTab&quot;:true,&quot;fPath&quot;:&quot;build/processor_board.elf&quot;,&quot;fProjectName&quot;:&quot;cansw_processor_canards&quot;,&quot;fPerformBuild&quot;:true,&quot;fDownload&quot;:true,&quot;fLoadSymbols&quot;:true}]}"/>
```
martendo, line 15: Why is this `build/processor_board.elf` and not something like `build/Debug/processor_board.elf`?

processor_board_Debug.launch
```diff
@@ -62,18 +61,18 @@
...
     <booleanAttribute key="org.eclipse.cdt.dsf.gdb.UPDATE_THREADLIST_ON_SUSPEND" value="false"/>
-    <intAttribute key="org.eclipse.cdt.launch.ATTR_BUILD_BEFORE_LAUNCH_ATTR" value="2"/>
+    <intAttribute key="org.eclipse.cdt.launch.ATTR_BUILD_BEFORE_LAUNCH_ATTR" value="0"/>
     <stringAttribute key="org.eclipse.cdt.launch.COREFILE_PATH" value=""/>
     <stringAttribute key="org.eclipse.cdt.launch.DEBUGGER_START_MODE" value="remote"/>
     <booleanAttribute key="org.eclipse.cdt.launch.DEBUGGER_STOP_AT_MAIN" value="true"/>
     <stringAttribute key="org.eclipse.cdt.launch.DEBUGGER_STOP_AT_MAIN_SYMBOL" value="main"/>
-    <stringAttribute key="org.eclipse.cdt.launch.PROGRAM_NAME" value="Debug/processor_board.elf"/>
-    <stringAttribute key="org.eclipse.cdt.launch.PROJECT_ATTR" value="processor_board"/>
+    <stringAttribute key="org.eclipse.cdt.launch.PROGRAM_NAME" value="build/processor_board.elf"/>
```
martendo, line 69: Same comment as above

martendo (commented):
The details of how this works go over my head so I have no comments there, but it appears to work (successful builds and tests) on my Arch Linux machine, so good job!

The only feedback on the changes themselves I have the knowledge to give is pedantry; none of this has noticeable impact, but nevertheless:
- Having `build` and `tests/build` in `.gitignore` is redundant
- There is no newline at the end of `.devcontainer/devcontainer.json`, `.gitignore`, `CMakePresets.json`, `cmake/gcc-arm-none-eabi.cmake`
- There are 2 extra newlines at the end of `cmake/stm32cubemx/CMakeLists.txt`

(I'm aware some of those files are generated, just calling it out in case something was missed.)

The combination of building and testing in a single script made sufficient sense to me. It could be nice to separate them because they are separate things but I personally don't think it would be necessary.

processor_board_Debug.launch
celery6: fixed in latest change

celery6:
> There is no newline at the end of `.devcontainer/devcontainer.json`, `.gitignore`, `CMakePresets.json`, `cmake/gcc-arm-none-eabi.cmake`

i just realized vscode's default .json file formatter doesn't include eof newline?? i dont think it's worth fixing (ie idk how) but that is really sad :(

martendo (approved):

Joe-Joe-Joe-Joe (dismissed):
As a user, the experience was pretty seamless, and the instructions worked.

The only thing I might add is the specific VScode command to load the dev container. In my case I just closed and reopened VScode hoping it would autodetect the dockerfile and ask me if I want to load it, and it did.

celery6 (dismissed):
i approve of @taisirhassan 's gtest setup

CMakePresets.json
taisirhassan: Honestly, even with the similar build configs, there’s no need to ditch the presets. They’re not causing any issues, and you never know when they might come in handy down the line. No point in overcomplicating it.

scripts/run.sh
taisirhassan: If confusion does come up later, maybe breaking it into two scripts would be worth revisiting. But for now, I’d keep it combined. It’s cleaner, and the overlap between the two workflows seems like more of a reason to combine than to separate.

.github/workflows/build.yml
```diff
@@ -1,17 +1,27 @@
...
+      - name: Install dependencies
+        run: |
+          sudo apt update
+          sudo apt install -y cmake ninja-build
+          wget -q https://github.com/xpack-dev-tools/arm-none-eabi-gcc-xpack/releases/download/v13.3.1-1.1/xpack-arm-none-eabi-gcc-13.3.1-1.1-linux-x64.tar.gz
+          tar -xzf xpack-arm-none-eabi-gcc-13.3.1-1.1-linux-x64.tar.gz
+          echo "xpack-arm-none-eabi-gcc-13.3.1-1.1/bin/" >> $GITHUB_PATH
+          
+      - name: Build firmware
+        run: ./scripts/run.sh build Debug
```
patrick-gu, line 26: do we need this workflow if there is a unit test one?

CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+    src/third_party/canlib/message/msg_gps.c
+    src/third_party/canlib/message/msg_recovery.c
+    src/third_party/canlib/message/msg_sensor.c
+    src/third_party/canlib/message/msg_state_est.c
+    src/third_party/canlib/stm32h7/stm32h7_can.c
+    src/third_party/canlib/util/can_rcv_buffer.c
+    src/third_party/canlib/util/can_tx_buffer.c
+    src/third_party/canlib/util/safe_ring_buffer.c
+    src/third_party/canlib/util/timing_util.c
+    src/third_party/printf/printf.c
```
patrick-gu, line 69: it would be nice to not have to list all of these off but I'm not sure how I would do that

README.md
```diff
@@ -1,16 +1,57 @@
...
+This project is not dependent on STM32CubeIDE.
+Code editing, unit testing, and building should be done in the devcontainer.
+Only running/debugging on target should be done in STM32CubeIDE.
+
+#### 1. Clone repo
+- Clone repo and initialize submodules: ```
+   git clone --recurse-submodules https://github.com/waterloo-rocketry/cansw_processor_canards ```
+
+#### 2. Edit, test, build in devcontainer
+The devcontainer contains the necessary environment setup for editing, unit testing, and building. Most software development work should occur here.
```
patrick-gu, line 22: If you need the build tools already to use with hardware, why use this? I imagine it's a bit slower. (I'm assuming there is a good reason, it just seems that this might be unnecessary?)

README.md
```diff
@@ -1,16 +1,57 @@
...
+#### 3. Run/debug in STM32CubeIDE
+STM32CubeIDE is preferred for flashing/debugging on hardware. NOTE: STM32CubeIDE is not set up to build this project. STM32CubeIDE is only used to run the build from the previous step.
+- Import the project into STM32CubeIDE (version 1.16.1 recommended): `File->Import...->Existing Projects into Workspace`
+- Build the project (previous step)
+- Use an ST-Link programmer to connect to processor board.
+- Use STM32CubeIDE run/debug as usual. Launch configurations are persisted in the `.launch` files in this repo.
+  - NOTE: since the project can't be built in STM32CubeIDE, auto-building before launch is turned off. You must manually build the project before launching if there are any changes.
+
+#### *ALTERNATIVE 3. Run/debug via Vscode STM32 extension*
+*Instead of using STM32CubeIDE, run/debug in vscode is possible using the STM32 vscode extension ([setup info here](https://community.st.com/t5/stm32-mcus/how-to-use-vs-code-with-stm32-microcontrollers/ta-p/742589) ). This was omitted from the devcontainer because usb passthrough into devcontainers is untrivial.*
```
patrick-gu, line 41: non-trivial

scripts/run.sh
```diff
@@ -0,0 +1,136 @@
...
+
+run_all() {
+  echo "Running full pipeline: Clean -> Build $1 and gtest -> Test"
+  clean_project "$1"
+  build_project "$1"
+  run_tests
+}
+
+show_help() {
+  echo "Usage: $0 {build|cleanbuild|test|clean|all|help} [target]"
```
patrick-gu, line 91: this script seems nice but I feel like it's a bit overkill

tests/CMakeLists.txt
```diff
@@ -0,0 +1,81 @@
...
+    mocks         # Mock library
+    proc_src      # Source code library
+    gtest_main    # GoogleTest main function
+)
+
+# Fetch GoogleTest
+include(FetchContent)
+FetchContent_Declare(
+    googletest
+    GIT_REPOSITORY https://github.com/google/googletest.git
```
patrick-gu, line 72: could this be a submodule? so it doesn't have to be fetched at build time

patrick-gu (commented):
I haven't checked this very carefully but generally everything seems fine to me, and if it works I don't think there are any major issues

scripts/run.sh
patrick-gu: could be easier to just make them all separate scripts

CMakeLists.txt
celery6: other option was to glob include everything (like the line above), then explicitly list exclusions. But theres more exclusions than inclusions for these 3rdparty libs so i think this is better (both options r annoying ngl)

README.md
celery6: use devcontainer to mandate a consistent unit test toolchain, cuz gcc doesnt exactly exist on windows. (ur right abt the actual hardware toolchain though)

scripts/run.sh
celery6: what specifically is overkill?

tests/CMakeLists.txt
celery6: @taisirhassan

.github/workflows/build.yml
celery6: oh i meant to raise that earlier - should we combine building the firmware binary into the unittests workflow? Rn the unit test one doesn't include building the firmware binary, it only builds for gtest.

But then, why is clang-format also a separate workflow when it coudl theoreticlaly be run with build/test?

celery6: googled for a min and yeah i think build firmware binary + unit tests + clangformat should each be actions in the same workflow. Thats faster too
ill make that change rn

CMakeLists.txt
patrick-gu: fair enough

README.md
patrick-gu: Is gcc required for unit tests?

scripts/run.sh
patrick-gu: Having this sort of command line interface for a script to run things. It might just be easier to just have a `build.sh`, `test.sh`, etc

README.md
celery6: some C compiler is required. scroll up commits to see the original setup, where we tried non-devcontainer using mingw for windows and gcc for other. this was extremely painful for both me and joe to set up, while devcontainers was instant working

scripts/run.sh
celery6: previously discussed here https://github.com/waterloo-rocketry/cansw_processor_canards/pull/43#discussion_r1903314422 idk whats easier

.github/workflows/build.yml
celery6: updated to be steps within a job. hope thats not against separation of concerns too much.. but its way faster to avoid setting up the action runner everytime. This seems like a fairly common setup for github actions (checks in multiple steps of 1 job)

scripts/run.sh
patrick-gu: Fair enough, as long as it's been considered

README.md
patrick-gu: Ok, I don't think it's really an issue, since it's easy to set up. I guess it just seems it could be flaky (i.e. if we don't know the specific issues) but this doesn't really matter.

I think from a software engineering perspective, it doesn't matter to have this be broadly compatible. We have the reproducible environment, and presumably we are never going to reuse it.

.github/workflows/build.yml
patrick-gu: Looks good! Thanks

patrick-gu:
@celery6 thanks for addressing my points. Sorry they were repetitive - I haven't been keeping track of this thread

celery6 (dismissed):

celery6 (approved):
LGTM!!!!!!!!

Armaan-Sengupta (approved):
lgtm

PR #49, opened 2025-01-02
timer header file and .ioc file configs

abdulmustafamohib:
-added header file for timer module
-changed configs in .ioc file to match last year's code
-timer frequency set to 0.1 ms


Joe-Joe-Joe-Joe:
I think `get_us` is kinda misleading if our resolution is only 100us. Is there a specific use case it is designed for? @celery6

celery6:
> I think `get_us` is kinda misleading if our resolution is only 100us. Is there a specific use case it is designed for? @celery6

@Armaan-Sengupta ?

Armaan-Sengupta:
Agreed it is confusing so we can get rid of it, actual us resolution would be nice though

Joe-Joe-Joe-Joe:
I mean if you want to implement 1us resolution, this is absolutely the correct PR to decide and implement it. But I'd again ask what the justification is?

Armaan-Sengupta:
As an example I have previously used it when debugging an SPI communication issue where the clock and data lanes where not synced correctly. However there is a very good chance it never gets used, so if we wan't to minimize the potential for wasted work, then sure, we can cross that bridge if we get to it.

src/drivers/timer/timer.h
```diff
@@ -0,0 +1,8 @@
+#ifndef TIMER_H
+#define TIMER_H
+
+float timer_get_us(void);
+
+float timer_get_ms(void);
```
celery6, line 6: add brief comments since this is for public usage. Ms from power up? From program startup? 
Also these should return rocketlib status code unless they definitely can't fail? in which case that should be acknoledged in comments too

celery6 (changes requested):

celery6:
i think if microseconds isn't actually a req in this project then we should skip implementing that. Timer resolution seems to be very project-specific so this wouldnt really improve code reuseability (or any other design benefits).

Armaan-Sengupta:
k, sounds good, lgtm.

celery6 (dismissed):

Redstone-ray (dismissed):

Redstone-ray (approved):

celery6:
force merging cuz pipelines broken

Issue #20, opened 2024-10-17
movella IMU module design


celery6:
See #5 

This module needs to coordinate with the polulu module as the interface to access an imu should be consistent across our project. see #12 for details

the movella IMU communicates via UART. this module needs to manage receiving raw sensor data from UART and do any decoding required to interpret the data.

Joe-Joe-Joe-Joe:
@Armaan-Sengupta do we have an interface library similar to what we had to the VN? Can you link/add it if you have it?

Armaan-Sengupta:
Some other ppl are working on the interface to the Movella, I haven't gotten to be in person yet. @celery6 knows who is working on it

Joe-Joe-Joe-Joe:
Ok i found it
https://www.movella.com/support/software-documentation
Its not hidden like the VN libraries lol

PR #50, opened 2025-01-11
Movella IMU Header & Submodule

alboainin:
This pull request will add the module header file, and the github submodule for the xsens-mti library that is used by this module. More information on the module on the firmware design document.


src/drivers/movella/movella.h
```diff
@@ -0,0 +1,63 @@
...
+#include "stm32h7xx_hal.h"
+#include "task.h"
+#include "xsens_constants.h"
+#include "xsens_mti.h"
+
+typedef struct {
+    float x;
+    float y;
+    float z;
+} Vector3f;
```
celery6, line 19: https://github.com/waterloo-rocketry/cansw_processor_canards/pull/44
added these project-wide vector type, pls use for consistency

src/drivers/movella/movella.h
```diff
@@ -0,0 +1,63 @@
...
+    Vector3f acc;
+    Vector3f gyr;
+    Vector3f euler;
+    Vector3f mag;
+    float pres;
+    float temp;
+    Quaternion quat;
+} MovellaData_t;
+
+bool movella_init(void);
```
celery6, line 50: any reason these functions all return bool instead of the rocketlib w_status_t?

celery6 (changes requested):

Joe-Joe-Joe-Joe (commented):
`movella_get_orientation` -> `movella_get_euler_angles` (the quaternion also represents orientation so this is somewhat ambiguous)
`movella_get_rate_of_turn` -> `movella_get_angular_vel` (this is my personal preference)

gdoc is lakcing detail, how are the individual extractors (get_accel, etc) meant to work with each other? Are they just wrappers for `get_imu_axes`? That is, will calling get_accel and then get_orientation read 2 separate packets and get their contents? Honestly I think that makes sense given any application that uses multiple outputs should really be unpacking a `MovellaData_t`. But then, what is the use case for these individual access functions?

alboainin:
@Joe-Joe-Joe-Joe initially I thought that we want to develop the IMUs public interface getters functions to be seperate, like a getter for acceleration, getter for pressure, etc... But in terms of design they will read the same packet and _might_ read the next packet, because, there will be a MovellaData_t instance that will be updated whenever a new package get recieved, and the getters function will just copy the latest instance to your variable. And the movella_get_imu_axes is exactly similar to the other getters function but instead of copying only one data type, it will copy multiple.

src/drivers/movella/movella.h
```diff
@@ -0,0 +1,49 @@
...
+w_status_t movella_get_imu_axes(MovellaData_t *out_data);
+
+w_status_t movella_get_utc_time(XsensUtcTime_t *out_utc);
+w_status_t movella_get_acceleration(vector3d_t *out_acc);
+w_status_t movella_get_rate_of_turn(vector3d_t *out_gyro);
+w_status_t movella_get_orientation(vector3d_t *out_euler);
+w_status_t movella_get_magnetometer(vector3d_t *out_mag);
+w_status_t movella_get_pressure(float *out_pres);
+w_status_t movella_get_temperature(float *out_temp);
+w_status_t movella_get_quaternion(quaternion_t *out_quat);
```
celery6, line 47: plss add brief comments indicating units of measurement it takes 3 sec and saves so much confusion for everyone

celery6 (changes requested):

src/drivers/movella/movella.h
```diff
@@ -0,0 +1,33 @@
...
+#include "drivers/uart/uart.h"
+
+typedef struct {
+    vector3d_t acc; // (x, y, z) m/s^2
+    vector3d_t gyr; // (x, y, z) rad/s
+    vector3d_t euler; // (x, y, z) deg
+    vector3d_t mag; // (x, y, z) arbitrary units, note: need to be converted to uT later
+    float pres; // Pa
+    float temp; // °c
+} MovellaData_t;
```
Joe-Joe-Joe-Joe, line 22: Caps

Joe-Joe-Joe-Joe (changes requested):
All the mti driver files you copied should go in `third_party/` (and rightly ought to provide reference to the original author)

Joe-Joe-Joe-Joe:
closes #20

celery6 (approved):
@alboainin 's changes look good to me . I also made some changes to clean this up pls review :pplslpsplspls

Joe-Joe-Joe-Joe (approved):
_our_ code? LGTM please god just let this milestone be done

The xsens library? Holy build warnings, Batman. I am not sure what the best policy is for 3rd party code that violates our warning settings. There are some bits of FATFS/SD code that also do this. We can probably bypass unused variable warnings on 3rd party code, since those are really only useful for detecting our own coding fuckups
image

PR #55, opened 2025-01-16
I2C-implementation

taisirhassan:


src/drivers/i2c/i2c.c
```diff
@@ -1 +1,267 @@
...
+ * @brief Error statistics for each bus
+ */
+static struct {
+    uint32_t timeouts; /**< Number of timeout errors */
+    uint32_t nacks; /**< Number of NACK errors */
+    uint32_t bus_errors; /**< Number of other bus errors */
+} error_stats[I2C_BUS_COUNT] = {0};
+
+// Forward declarations for internal functions
+static void i2c_transfer_complete_callback(I2C_HandleTypeDef *hal_handle, bool success);
```
celery6, line 49: is this the right function signature for hal i2c callbacks? where does it have the succcess parameter
image
i thought this was the function signature but idk never looked that closely

src/drivers/i2c/i2c.c
```diff
@@ -1 +1,267 @@
...
+        ); // Clean up the mutex if the semaphore creation failed to avoid resource leaks
+        // TODO: Add logging when logger is available
+        return W_FAILURE;
+    }
+    // Register the transfer completion callback with the HAL driver for the bus instance to handle
+    // transfer completion events
+    HAL_I2C_RegisterCallback(
+        handle->hal_handle, HAL_I2C_MASTER_TX_COMPLETE_CB_ID, (void *)i2c_transfer_complete_callback
+    );
+    HAL_I2C_RegisterCallback(
```
celery6, line 100: these registercallback calls can fail, u should check the outcomes

src/drivers/i2c/i2c.c
```diff
@@ -1 +1,267 @@
...
+// STM32 HAL I2C Event Interrupt Handlers
+void HAL_I2C_MasterTxCpltCallback(I2C_HandleTypeDef *hi2c) {
+    i2c_transfer_complete_callback(hi2c, true);
+}
+
+void HAL_I2C_MasterRxCpltCallback(I2C_HandleTypeDef *hi2c) {
+    i2c_transfer_complete_callback(hi2c, true);
+}
+
+void HAL_I2C_ErrorCallback(I2C_HandleTypeDef *hi2c) {
```
celery6, line 265: these do nothing if u used the hal registercallback thing. I think these are if u want to override the default weak callbacks? but since we explicitly used registercallback, it will use the registered callback instead of any weak overrides

src/drivers/i2c/i2c.c
celery6: nvm, see comment at bottom of this file :FACE:

src/drivers/i2c/i2c.c
```diff
@@ -1 +1,267 @@
...
+
+/**
+ * @brief Callback function for transfer completion
+ *
+ * Called from ISR context when a transfer completes or fails.
+ *
+ * @param[in] hal_handle HAL I2C handle that completed
+ * @param[in] success Whether the transfer was successful
+ */
+static void i2c_transfer_complete_callback(I2C_HandleTypeDef *hal_handle, bool success) {
```
celery6, line 223: if this is called in response to both rx and tx complete, it won't clash right 🤔 
both rx/tx are protected by 1 mutex? just wanted to double check this logic

celery6 (changes requested):

src/drivers/i2c/i2c.c
taisirhassan: yeah the mutex makes it safe - it prevents a new transfer from starting until the previous one is fully done. so when the callback runs, we know it can only be for one specific transfer. Doesn't matter if it's RX or TX since they can't happen at the same time anyway lol.

celery6: ohyeah skull imagine not being fullduplex smh i2c

Core/Src/main.c
celery6, line 1: can u make sure the .ioc file is configured for all the new i2c connections, and remove all last year's confgs. And then init them all in main.c. There will be an improved init routine later, but for now just stick all inits in main.c so we know they exist

tests/unit/i2c_test.cpp
```diff
@@ -0,0 +1,185 @@
...
+        }
+        // For all other cases (including the zero-time "clear" call and mutex), return pdTRUE immediately.
+        return pdTRUE;
+    };
+
+    uint8_t data[4];
+    w_status_t status = W_SUCCESS;
+
+    // Start the read operation in a separate thread.
+    std::thread readThread([&]()
```
celery6, line 126: do not mock threads in unit tests :sob: 
this is like integration testing idk. can u comment out till we conclude on a way to test freertos properly?

celery6 (changes requested):
hi

celery6 (changes requested):

Joe-Joe-Joe-Joe:
I tested this IRL when I tested #147 , it seems to work (at least on i2c4)

.devcontainer/devcontainer.json
```diff
@@ -21,6 +21,10 @@
     "postStartCommand": {
         // Enable git tab-autocompletion in terminal
         "autocomplete": "echo \"source /usr/share/bash-completion/completions/git\" >> ~/.bashrc",
+
+        // Use vscode as default editor cuz nano/vim arent installed and ppl can install those by preference
```
celery6, line 25: delete this change its old

Core/Src/main.c
```diff
@@ -47,6 +51,11 @@
 /* Private define ------------------------------------------------------------*/
 /* USER CODE BEGIN PD */
 #define DEFAULT_STACKDEPTH_WORDS 128 * 4
+
+#define DEVICE_ADDR (0x6B)   // LSM6DSO address (0x6B = 01101011)
+#define WHO_AM_I_REG (0x0F)  // WHO_AM_I register
```
celery6, line 56: delete

Core/Src/main.c
```diff
@@ -165,6 +180,42 @@ int main(void)
   /* USER CODE END 3 */
 }
 
+void I2CTask(void *argument)
```
celery6, line 183: delete all testing code

celery6 (changes requested):

celery6 (approved):
hopefully it works :FACE:

PR #61, opened 2025-01-18
Fix cubeide

celery6:
pls someone else test if u are able to press the run/debug buttons in stm32cubeide from this branch

i hate stm32cubeide

launching from stm32cubeide somehow got broken along the way. now it should correctly detect the launch config, and also the ioc file.


.cproject
```diff
@@ -0,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
```
Hany25, line 1: Ran the code file in STM32Cube; is working as intended.

Hany25 (approved):
To approve pr

PR #65, opened 2025-01-20
Gpio implementation

celery6:
tested on processor board by calling various gpio functions in defaulttask - all gpio functions work as expected AND I UPDATED THE IOC FILE CORRECTLY.

also added unit tests, HAL gpio mocks, freertos semphr mocks


tests/unit/gpio_test.cpp
```diff
@@ -0,0 +1,198 @@
...
+TEST_F(GpioTest, InitCreatesMutexes) {
+    // Arrange
+    xSemaphoreCreateMutex_fake.return_val = (SemaphoreHandle_t)1;
+
+    // Act
+    w_status_t status = gpio_init();
+
+    // Assert
+    EXPECT_EQ(status, W_SUCCESS); // Ensure initialization succeeds
+    EXPECT_EQ(xSemaphoreCreateMutex_fake.call_count, GPIO_PIN_COUNT); // Check mutex created for each pin
```
celery6, line 37: ngl idk how valid this method of testing using mocks is. This is quite whiteboxing it cuz if smth abt the semaphore implementation changes then the tests break despite it still working so uhhh ???

But creating actual fake implementations of semaphore/other freertos behaviour would be so ridiculously excessive compared to just counting how many times the funcs are called

celery6:
> I didn't notice it in the first tests PR but the header files here use `#define`s without leading underscores, disagreeing with the coding standard — is that intentional?

i'm assuming ur referring to the standard on this page? https://docs.waterloorocketry.com/general/standards/embedded-coding-standard.html
interesting i just noticed this contradicts the BARR standard ... and also the C11 standard itself which says identifiers starting with a underscore+capital letter are reserved. I'll change the docs site

taisirhassan (approved):
LGTM!!!!

celery6:
merged with main, hopefulyly that didnt break the .ioc ....

taisirhassan (approved):

PR #66, opened 2025-01-21
Add logger header file

martendo:
See [Firmware Design &sect; Logger](https://docs.google.com/document/d/1OokvdQ8c74xzt_znfmBUrUzJ5vtqvndMz0l9xxN98dY/edit?tab=t.0#heading=h.zj730k0v797) for module design.

### DIFF
--- a/src/application/logger/log.h
+++ b/src/application/logger/log.h
@@ -1,0 +1,65 @@
+#ifndef _LOG_H
+#define _LOG_H
+
+#include "rocketlib/include/common.h"
+
+// TODO: Determine optimal numbers for these
+/* Size of a single buffer (bytes) */
+#define LOG_BUFFER_SIZE 16384
+/* Size of each message region in text buffers (bytes) */
+#define MAX_TEXT_MSG_LENGTH 256
+/* Size of each message region in data buffers (bytes) */
+#define MAX_DATA_MSG_LENGTH 64
+/* Number of message regions in a single text buffer */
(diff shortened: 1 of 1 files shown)


celery6:
added comments in the google doc and so did bestie Patrick!!

src/application/logger/log.h
```diff
@@ -0,0 +1,97 @@
...
+ * Version number to identify post-flight how to parse data log messages.
+ * Increment this value when making an incompatible change to log_data_type_t or log_data()'s
+ * formatting of messages.
+ *
+ * Deprecated values: none
+ */
+#define LOG_DATA_FORMAT_VERSION 1
+
+/* Magic number to encode into log_data_type_t values: "DL" encoded as a little-endian 16-bit int */
+#define LOG_DATA_MAGIC 0x4c44
```
celery6, line 41: random personal opinion but i think these magic cookies should look meaningful in hex if that's probably how we would read the raw file?
imagine the possibilities... 0xF00DBEEF .. 0x00FACE11

celery6 (approved):
LGTM!!!!!!!!! (Lets Good This Me!)

Redstone-ray (approved):

src/application/logger/log.h
martendo: I think it's easier to pick out markers if they're in ASCII, because many values are not valid ASCII characters and so they stand out in that way. I typically use something like `hexdump -C` (I think this is common, or at least good hex editor programs should also have a way to show both hex and ASCII), with which it should be easy to see these ASCII characters.

PR #70, opened 2025-01-22
added a struct and function to get constants for voltage and current …

jerryjcr:
…conversions

### DIFF
--- a/src/drivers/adc/adc.h
+++ b/src/drivers/adc/adc.h
@@ -3,11 +3,22 @@
 
 #include "rocketlib/include/common.h"
 #include <stdbool.h>
+#include <stdint.h>
 
 typedef enum {
     PROCESSOR_BOARD_VOLTAGE = 0,
     ADC_CHANNEL_COUNT
 } adc_channel_t;
+
+typedef struct {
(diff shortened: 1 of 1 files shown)


src/drivers/adc/adc.h
```diff
@@ -3,12 +3,23 @@
...
     PROCESSOR_BOARD_VOLTAGE = 0,
     ADC_CHANNEL_COUNT
 } adc_channel_t;
 
+typedef struct {
+    double ADC_MAX_COUNTS;
+    double ADC1_VREF;
+    double R_SENSE;
+    double INA180A3_GAIN;
+} adc_constants_t;
```
celery6, line 18: if we added more adc channels in the future would this struct just grow indefinitely..?
I imagined this would be per channel 🤔

jerryjcr: Will we just have a different struct for every channel then? Should I just rename the struct to something like processor_voltage_constants_t?

Joe-Joe-Joe-Joe:
Why does this exist and why is it in a struct? I guess you don't want to use 5 `#define`s? But I don't think this is any more extensible, it just seems tedious for what it is.

PR #76, opened 2025-01-26
Created SD Card header file

Wehbucci:
This PR introduces the header interface of the FATFS wrapper funuction, used to interact with an onboard SD card.


src/drivers/sd_card/sd_card.h
```diff
@@ -0,0 +1,18 @@
+#ifndef SD_CARD_H
+#define SD_CARD_H
+
+#include <fatfs.h>
+
+FRESULT sdCardInit(void);
```
celery6, line 6: all of these should return the teamwide rocketlib status code instead of FRESULT. See other prs for examples, and https://docs.waterloorocketry.com/general/standards/embedded-coding-standard.html#error-handling

celery6: btw the correct include path for that is `#include "rocketlib/include/common.h`

src/drivers/sd_card/sd_card.h
```diff
@@ -0,0 +1,18 @@
...
+
+FRESULT fileRead(char *fileName, void *buffer, UINT bufferSize, UINT *bytesRead);
```
celery6, line 8: this public interface shouldn't have any fatfs-specific types like UINT, cuz the whole point of this is to allow anyone to use it without requiring knowledge of what a UINT is.

src/drivers/sd_card/sd_card.h
celery6, line 1: use `snake_case` for naming (everything lowercase, separated by underscore) as per teamwide coding guideline
https://docs.waterloorocketry.com/general/standards/embedded-coding-standard.html#coding-style

src/drivers/sd_card/sd_card.h
```diff
@@ -0,0 +1,18 @@
+#ifndef SD_CARD_H
+#define SD_CARD_H
+
+#include <fatfs.h>
```
celery6, line 4: after addressing the other comments, i think this can be removed from this header file and only need it in the .c implementation file? it's better to reveal as little as posisble in this header file cuz its public

src/drivers/sd_card/sd_card.h
celery6, line 1: add brief comments for each function declared here - we can't assume ppl will make the correct assumptions when using the functions. See other prs for example of comment detail level

celery6 (changes requested):
could you update the design doc to have a few notes about the new init and status functions? how will those work? what do they do? what kind of status can sd card return?

celery6:
the design doc vaguely mentions having a mutex.. somewhere? how will that work? u should add some notes abt that before implementing it. The purpose of this is to reduce plot holes by thinking ahead

martendo (commented):
@celery6 already hit everything I would've said

celery6:
lgtm!!

celery6 (approved):

martendo (approved):
Function parameter names are still not snake case but I don't know how strict we should be &mdash; LGTM otherwise

Issue #72, opened 2025-01-23
health checks/watchdog implementation


celery6:
may be nicer to split this implementation into 2 steps (2 pull reqs as well), first for just health checks, then with watchdog

PR #77, opened 2025-01-27
#72 health checks implementation

kimmyhoang:
Monitors for over current and send CAN messages:
  - interfaces with ADC and CAN handler 
- `get_adc_current` ensures `adc_get_value` is returning success
- `health_check_task` checks for overcurrent and sends CAN message

Implements a software watchdog check:
- Tasks register with timeout values
- Tasks must "kick" watchdog periodically
- Sends timeouts over CAN message for unresponsive tasks using `check_watchdog_tasks`

health_check_task runs checks every 1000ms in an infinite loop


Joe-Joe-Joe-Joe:
Am I missing something or do you want to merge this before implementing the building of the CAN message?

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,74 @@
...
+        log_text("HealthCheck", "Overcurrent detected: %dmA", adc_current_mA);
+      } else if (adc_current_mA < constants.MIN_CURRENT_mA) {
+        log_text("HealthCheck", "Undercurrent detected: %dmA", adc_current_mA);
+      } else {
+        log_text("HealthCheck", "Current draw nominal: %dmA", adc_current_mA);
+      }
+
+      // TODO: monitor watchdog tasks
+    }
+    vTaskDelayUntil(&lastWakeTime, pdMS_TO_TICKS(1000));
```
celery6, line 72: this function has a return value that u should check to make sure the time didn't overrun past 1000ms
(google it)

celery6: idk if thats smth for implementation part2? add a todo at least

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,74 @@
...
+  for (;;) {
+    if (get_adc_current(&adc_current_mA) == W_SUCCESS) {
+      if (adc_current_mA > constants.MAX_CURRENT_mA) {
+        can_msg_t msg;
+        uint8_t current_data[2] = {
+            (adc_current_mA >> 8) & 0xFF, // High byte
+            adc_current_mA & 0xFF         // Low byte
+        };
+
+        // TODO: build_msg()
```
celery6, line 58: there's going to be a new CAN err msg builder later so add that as a //TODO for overcurrent errs.

But we also want to log every 5v current measurement which u can implement now since that's done via canlib 
image

celery6: same as last year i think^

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,74 @@
...
+      (uint16_t)((((float)adc_value) / constants.ADC_MAX_COUNTS) *
+                 constants.ADC_VREF) *
+      1000;
+  *adc_current_mA =
+      (uint16_t)((voltage_mV / constants.INA180A3_GAIN) / constants.R_SENSE);
+
+  return W_SUCCESS;
+}
+
+w_status_t health_check_init(void) {
```
celery6, line 34: add this init to Core/Src/main.c. also register the task with freertos, copy last year's conventions for that (also in main.c)
image

celery6 (changes requested):
make sure u frequently merge main into ur branch and stay uptodate w the readme updates

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,216 @@
...
+
+    for (;;) {
+        health_check_exec();
+
+        vTaskDelayUntil(&lastWakeTime, pdMS_TO_TICKS(TASK_DELAY_MS));
+        TickType_t now = xTaskGetTickCount();
+
+        if ((now - lastWakeTime) > pdMS_TO_TICKS(TASK_DELAY_MS)) {
+            // to do log this error?
+        }
```
celery6, line 214: are u sure this is doing what u intended to? It looks like this is trying to check if the exec function took too long?
cuz after the vtaskdelayuntil, the next line of code runs immediately. It will always be within 5ms of execution time. so thats not very helpful and is giving a false sense of security imo.

the "correct" way to check if a task overran the taskdelayuntil() is to use xtaskdelayuntil() (prepended with x instead of v). Except we can't use that because it's in a newer version of freertos than ours so yeah too bad

https://forums.freertos.org/t/vtaskdelayuntil-vs-v-xtaskdelayuntil/19085

tests/mocks/task.c
```diff
@@ -1,8 +1,8 @@
...
-DEFINE_FAKE_VOID_FUNC(vTaskDelay, TickType_t);
-DEFINE_FAKE_VALUE_FUNC(TickType_t, xTaskGetTickCount);
-DEFINE_FAKE_VALUE_FUNC(BaseType_t, xTaskDelayUntil, TickType_t*, TickType_t);
-DEFINE_FAKE_VOID_FUNC(vTaskDelayUntil, TickType_t*, TickType_t);
+// DEFINE_FAKE_VALUE_FUNC(TaskHandle_t, xTaskGetCurrentTaskHandle);
+// DEFINE_FAKE_VALUE_FUNC(BaseType_t, xTaskGetSchedulerState);
+// DEFINE_FAKE_VOID_FUNC(vTaskDelay, TickType_t);
+// DEFINE_FAKE_VALUE_FUNC(TickType_t, xTaskGetTickCount);
+// DEFINE_FAKE_VALUE_FUNC(BaseType_t, xTaskDelayUntil, TickType_t*, TickType_t);
+// DEFINE_FAKE_VOID_FUNC(vTaskDelayUntil, TickType_t*, TickType_t);
```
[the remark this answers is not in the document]
celery6: i recently added info abt this in the practical coding advice doc 
image

src/application/can_handler/can_handler.c
```diff
@@ -5,6 +5,14 @@
...
 #define BUS_QUEUE_LENGTH 16
 
+bool check_board_need_reset(const can_msg_t *msg) {
+    // hard coded the implementation for now as it kept giving me some errors
+    // TODO : ACTUAL IMPLEMENTATION OF THIS FUNCTION
+    // understand how to later
+    (void)msg;
+    return false; // should never reach there anyway
+}
+
```
celery6, line 15: this function is provided by canlib so delete this

src/application/flight_phase/flight_phase.c
```diff
@@ -11,6 +11,8 @@
 // See the flowchart in the design doc for more context on these
 #define ACT_DELAY_MS 9000 // Q - the minimum time after launch before allowing canards to actuate
 #define FLIGHT_TIMEOUT_MS 50000 // K - the approximate time between launch and apogee
+#define ACT_STATE_ON 1
+#define ACT_STATE_OFF 0
```
celery6, line 15: these types are also provided by canlib, so delete these

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,216 @@
...
+#include "printf.h"
+#include "task.h"
+
+#define TASK_DELAY_MS 1000
+#define ADC_VREF 3.3f
+#define R_SENSE 0.033f
+#define INA180A3_GAIN 100.0f
+#define MAX_CURRENT_mA 400
+#define MAX_WATCHDOG_TASKS 10
+#define E_WATCHDOG_TIMEOUT 0x81 //check with christine if this works
```
celery6, line 16: this is fine as a placeholder, but we will have to consolidate project-wide error bits later so just add a TODO comment that this is a placeholder.

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,216 @@
...
+            uint32_t ticks_elapsed = pdMS_TO_TICKS((uint32_t)time_elapsed); 
+
+            if (!watchdog_tasks[i].is_kicked && (ticks_elapsed > watchdog_tasks[i].timeout_ticks)) {
+                // TODO: LOGGING ERROR
+
+                can_msg_t msg = {0};
+                if (false == build_general_board_status_msg(
+                                 PRIO_HIGH, (uint16_t)current_time, E_WATCHDOG_TIMEOUT, i, &msg
+                             )) { 
+                                // figure this out what goes here, do i even log an error? lol
```
celery6, line 150: log as much as possible cuz why not

src/application/flight_phase/flight_phase.c
panthpatel2016: where are they? I keep getting build errors because they are not defined 
image

src/application/can_handler/can_handler.c
panthpatel2016: same issue, cant build without having this, how do i fix that?

src/application/health_checks/health_checks.c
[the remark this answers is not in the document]
panthpatel2016: changed accordingly

tests/mocks/task.c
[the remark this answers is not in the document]
panthpatel2016: changed it back, mb

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,216 @@
...
+    }
+
+    // Check if we have space for a new task
+    if (num_watchdog_tasks >= MAX_WATCHDOG_TASKS) {
+        // TODO: LOGGING ERROR
+        return;
+    }
+
+    // Check if the task is already registered
+    for (uint8_t i = 0; i < num_watchdog_tasks; i++) {
```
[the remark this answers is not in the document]
panthpatel2016: done!

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,216 @@
...
+        if (watchdog_tasks[i].task_handle == task_handle) {
+            
+            //TODO: what should we do with watchgdog_tasks[i].is_kicked here?
+
+            // Update the timeout if the task already exists 
+            watchdog_tasks[i].timeout_ticks = timeout_ticks;
+            return;
```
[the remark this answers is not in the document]
panthpatel2016: oh I see, done!

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,216 @@
...
+            if (false == build_general_board_status_msg(
+                             PRIO_LOW, (uint16_t)ms, E_NOMINAL, adc_current_mA, &msg
+                         )) {
+                return W_FAILURE;
+            }
+        }
+
+        status |= can_handler_transmit(&msg);
+
+        status |= check_watchdog_tasks();
```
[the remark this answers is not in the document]
panthpatel2016: that was a mistake, thanks for pointing it out. changed accordingly

src/application/health_checks/health_checks.c
panthpatel2016: makes sense

src/application/flight_phase/flight_phase.c
celery6: canlib is a git submodule so you need to update git submodules if u didn't. 
image
the command is `git submodule update --init --recursive`. u should do this whenever u merge with main in case someone updated a submodule.

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,196 @@
...
+w_status_t get_adc_current(uint32_t *adc_current_mA) {
+    w_status_t status = W_SUCCESS;
+    uint32_t adc_value;
+
+    status |= adc_get_value(PROCESSOR_BOARD_VOLTAGE, &adc_value, TASK_DELAY_MS);
+    if (status != W_SUCCESS) {
+        return status;
+    }
+
+    uint32_t voltage_mV = ((float)(adc_value) / ADC_MAX_COUNTS) * ADC_VREF *
```
Joe-Joe-Joe-Joe, line 40: why do you use uint32_t as intermediate representation?

I feel like it would be better to define a single conversion factor such as `CONV_ADC_COUNTS_TO_CURRENT_mA` in terms of the defines above, and then apply a single multiplication. Only need to cast at the start and end.

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+        status |= can_handler_transmit(&msg);
+    }
+
+    return status;
+}
+
+w_status_t health_check_init(void) {
+    for (uint32_t i = 0; i < MAX_WATCHDOG_TASKS; i++) {
```
Joe-Joe-Joe-Joe, line 79: this is equivalent to writing `static watchdog_task_t watchdog_tasks[MAX_WATCHDOG_TASKS] = {0};`
even if you wanted to initialize with nonzero values, our standard forbids uninitialized memory definitions, so you need to add static initialization regardless

Joe-Joe-Joe-Joe (changes requested):

src/application/health_checks/health_checks.c
panthpatel2016: This was implemented before I took over the PR, should I ask the person who worked on it before for maybe a better idea of the reason?

On the other hand, I agree there doesn't seem to be a good reason to get a value in one form just for it to be casted again. I was wondering it could be related to the comment left before hand where some values lost precision for no good reason?

src/application/health_checks/health_checks.c
panthpatel2016: changed accordingly.
just wondering, that would mean the health_check_init(); is basically useless, right? 
as all the initialisations would have to be static and global for all related functions to use them in healtchecks.c 
technically all of the functions can work without ever init ing? or am i getting something wrong

src/application/health_checks/health_checks.c
Joe-Joe-Joe-Joe: @celery6

celery6: the previous comment was literally a todo to fix this (seems like u interpreted it as a justification)

celery6: just re-do this calculation yourself using ur own knowledge. im guessing this was copypasted from a pic18 code that doesn't use floats

celery6 (changes requested):
image
ur 81 commits behind main. update ur main and update this branch. 
i think u missed a logger update so u need to update all the `log_text` because it has a timeout parameter now

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,201 @@
...
+        status |= can_handler_transmit(&msg);
+    }
+
+    return status;
+}
+
+// watchdog initiailsations
+static watchdog_task_t watchdog_tasks[MAX_WATCHDOG_TASKS] = {0};
+static uint32_t num_watchdog_tasks = 0;
+
```
celery6, line 80: why are these 2 variables declarations in the middle of the file

in 99% of cases, its standard to declare all variables/typedefs at the top of ur file, then write private functions, then write public functions

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,201 @@
...
+    can_msg_t msg = {0};
+
+    status |= timer_get_ms(&current_time);
+    if (status != W_SUCCESS) {
+        log_text("health_checks", "timer_get_ms failure");
+        return status;
+    }
+
+    for (uint32_t i = 0; i < num_watchdog_tasks; i++) {
+        if (watchdog_tasks[i].task_handle != NULL) {
```
celery6, line 153: its literally impossible for task_handle == null because you check that in the register function...so remove this `if` block cuz imo its redundant
or at least add the `else` to log failure if u think this case may occur

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,201 @@
...
+                if (false == build_general_board_status_msg(
+                                 PRIO_HIGH, (uint16_t)current_time, E_WATCHDOG_TIMEOUT, i, &msg
+                             )) {
+                    log_text("health_checks", "E_WATCHDOG_TIMEOUT status message error");
+                    return W_FAILURE;
+                }
+
+                status |= can_handler_transmit(&msg);
+            }
+            // resetting for next check
```
celery6, line 169: is last_kick_timestamp supposed to reset here too

tests/CMakeLists.txt
```diff
@@ -283,3 +283,11 @@ add_test_group(test_sd_card
     # src file(s)
     ${PROC_SRC_DIR}/drivers/sd_card/sd_card.c
 )
+
+add_test_group(test_health_check
+    # test file(s)
+    ${UNIT_TESTS_DIR}/health_checks_test.cpp
+    # src file(s)
+    ${PROC_SRC_DIR}/application/health_checks/health_checks.c
+    ${PROC_SRC_DIR}/application/logger/log.c
```
celery6, line 292: why add logger? ur not testing logger? create FFF fakes for log functions

tests/unit/health_checks_test.cpp
```diff
@@ -0,0 +1,247 @@
...
+// Mocked global variables
+static float timer_ms_value_mock;
+static uint32_t adc_value_mock;
+
+}
+
+DEFINE_FFF_GLOBALS;
+
+// Mocked functions for helping with unit testing 
+// check if having extra static functions is good or not for unit testing
```
celery6, line 38: helper functions are great and often required to make unit testing do what u need

celery6: though the helper functions should be orders of magnitude simpler than whatveer ur testing.. cuz otherwise ur also testing quality of helpers :skull 

so these helpers are good

panthpatel2016: cool makes sense, I was too sure about how complicated my functions were. Thanks

tests/CMakeLists.txt
panthpatel2016: oop, changed it now

src/application/health_checks/health_checks.c
panthpatel2016: makes sense, i removed that check as we would never reach there, therefore no need for logging too

src/application/health_checks/health_checks.c
panthpatel2016: oh my bad, i moved things around to organise my thoughts and left them in the middle 😭. they were orignially declared at the top
changed it now

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,196 @@
...
+
+    return W_SUCCESS;
+}
+
+w_status_t check_current(void) {
+    uint32_t adc_current_mA;
+    w_status_t status = W_SUCCESS;
+
+    status |= get_adc_current(&adc_current_mA);
+    if (status == W_SUCCESS) {
```
martendo, line 52: BARR says constant should be on the left of `==`:
```suggestion
    if (W_SUCCESS == status) {
```

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,196 @@
...
+            watchdog_tasks[i].is_kicked = true;
+            watchdog_tasks[i].last_kick_timestamp = current_time;
+            return W_SUCCESS;
+        }
+    }
+    return W_FAILURE;
+}
+
+w_status_t watchdog_register_task(TaskHandle_t task_handle, uint32_t timeout_ticks) {
+    if (task_handle == NULL || timeout_ticks == 0) {
```
martendo, line 106: constant on left and brackets as per BARR:
```suggestion
    if ((NULL == task_handle) || (0 == timeout_ticks)) {
```

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,196 @@
...
+        log_text(0, "health_checks", "invalid arguments into watchdog register");
+        return W_FAILURE;
```
martendo, line 108: we have a status value for this!
```suggestion
        return W_INVALID_PARAM;
```

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,196 @@
...
+
+            status |= can_handler_transmit(&msg);
+        }
+        // resetting for next check
+        watchdog_tasks[i].is_kicked = false;
+    }
+    return status;
+}
+
+w_status_t health_check_exec() {
```
martendo, line 171: does this function need to be exposed to other modules? should we make it `static`?

src/application/health_checks/health_checks.c
```diff
@@ -0,0 +1,196 @@
...
+    watchdog_tasks[num_watchdog_tasks].is_kicked = true;
+    watchdog_tasks[num_watchdog_tasks].last_kick_timestamp = current_time;
+    watchdog_tasks[num_watchdog_tasks].timeout_ticks = timeout_ticks;
+
+    num_watchdog_tasks++; // incriminent the watchdog task count for future ref
+
+    return status;
+}
+
+w_status_t check_watchdog_tasks(void) {
```
martendo, line 138: does this function need to be exposed to other modules? should we make it `static`?

martendo (changes requested):
I only have some nitpicks, otherwise looks pretty good!!

celery6 (approved):
LGMTMMTMMMMMMN!!!!!!!!

martendo (approved):

PR #79, opened 2025-01-29
Flight phase module design

### LINKED ISSUE #8: flight phase module design
see #5 

multiple people can contribute on this

Responsible for detecting flight phases. Also possibly taking action when certain phases are hit, like notifying controller maybe? Fairly simple module, but also really critically important to be accurate and not mess up!

Components:
- Task that continuously checks and updates current flight phase.

Interface:
- FlightPhase_Get_Current(). Return current flight phase, maybe? See point below
- FlightPhase_Send_Event(). Allow other modules to notify of relevant events, notably CAN for injector valve open event and possibly others

Things to figure out
- This module really smells like it should be explicitly an event-driven state machine? Last year’s was a large if-else chain, almost like a input-driven state machine. Problem: difficult to prove that works. If we can afford to, it would be amazing to redesign this as an actual state machine because this improves reliability, allowing formally verifying it, clear test cases, avoid race conditions, etc
- Testing: if a state machine, there’s steps to actually formally verify its validity. Google it. I have no idea if this is really practical, but looks very interesting to learn at least
- Last year’s flight phase had extensionAllowed() as a public interface, but maybe this isn’t entirely accurate? Ex., if in coast phase but something else causes abort, means we’re not actually allowed to actuate canards. Should this module just publish flight phase as the raw enum instead for other modules to interpret further?
- How detect recovery/descent better? There was talk about CAN msg from recovery?
instead of polling millis_() every loop, what if this used a software timer to trigger timed phases? Theoretically more precise since polling millis_() leaves variation Possibly less reliable though, since if you miss the timer completion ping then it's doomed.

Joe-Joe-Joe-Joe:
Procrastination moment
Closes #8 

[https://docs.google.com/document/d/1OokvdQ8c74xzt_znfmBUrUzJ5vtqvndMz0l9xxN98dY/edit?tab=t.0#heading=h.x7g2zix1yybt](https://docs.google.com/document/d/1OokvdQ8c74xzt_znfmBUrUzJ5vtqvndMz0l9xxN98dY/edit?tab=t.0#heading=h.x7g2zix1yybt)

- Now event driven using software timers
- States and transitions based on last year's
  - We had a weird velocity recovery trigger
  - How we detect recovery is subject to change, I can add an event to force change to recovery (and boost for that matter) if that is desired functionality

I've never designed a module before please go easy on me

### DIFF
--- a/src/application/flight_phase/flight_phase.h
+++ b/src/application/flight_phase/flight_phase.h
@@ -1,12 +1,7 @@
 #ifndef FLIGHT_PHASE_H
 #define FLIGHT_PHASE_H
 
-#include "FreeRTOS.h"
-#include "queue.h"
-#include "timers.h"
 #include "rocketlib/include/common.h"
-#include <stdbool.h>
-#include <stdint.h>
 
 /**
  * Enum representing phase of flight (state machine state)
--- a/src/application/flight_phase/flight_phase.h
+++ b/src/application/flight_phase/flight_phase.h
@@ -6,8 +6,7 @@
 /**
  * Enum representing phase of flight (state machine state)
  */
-typedef enum
-{
+typedef enum {
     STATE_PAD,
     STATE_BOOST,
     STATE_COAST,
@@ -18,8 +17,7 @@ typedef enum
 /**
  * Enum representing a state transition event
  */
-typedef enum
-{
+typedef enum {
     EVENT_INJ_OPEN,
     EVENT_BOOST_ELAPSED,
     EVENT_COAST_ELAPSED,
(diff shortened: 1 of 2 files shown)


Redstone-ray (approved):

celery6 (changes requested):
comments in gogle doc

src/application/flight_phase/flight_phase.h
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * Task to execute the state machine itself. Consumes events and transitions the state
+ */
+void flight_phase_task(void *args);
+
+/**
+ * Returns the current state of the state machine
+ * Not ISR safe
+ */
+flight_phase_state_t flight_phase_get_state();
```
martendo, line 46:
```suggestion
flight_phase_state_t flight_phase_get_state(void);
```

martendo (changes requested):

celery6 (approved):
left a random thought abt timers in the doc
lgtm!!!!!!!!!

PR #105, opened 2025-02-05
uart_write implementation

samcyx:


src/drivers/uart/uart.c
```diff
@@ -1 +1,89 @@
...
+    if (channel >= UART_CHANNEL_COUNT || uart_channel_map[channel].handle == NULL)
+    {
+        status = W_INVALID_PARAM; // Invalid parameter(s)
+        return status;
+    }
+    if (xSemaphoreTake(uart_channel_map[channel].write_mutex, timeout) != pdTRUE)
+    {
+        return W_IO_TIMEOUT; // Could not acquire the mutex in the given time
+    }
+    HAL_StatusTypeDef transmit_status = HAL_UART_TRANSMIT_IT(uart_channel_map[channel], data, len);
```
celery6, line 71: image
the firmware build gives this warning :thinking:

src/drivers/uart/uart.c
```diff
@@ -1 +1,89 @@
...
+// Returns the status of data transfer
+w_status_t uart_write(uart_channel_t channel, const uint8_t *data, uint8_t len, uint32_t timeout)
+{
+    w_status_t status = W_SUCCESS;
+    if (channel >= UART_CHANNEL_COUNT || uart_channel_map[channel].handle == NULL)
+    {
+        status = W_INVALID_PARAM; // Invalid parameter(s)
+        return status;
+    }
+    if (xSemaphoreTake(uart_channel_map[channel].write_mutex, timeout) != pdTRUE)
```
celery6, line 67: you forgot to give back this mutex once the function is finished
(this functionality is described in the design doc)

src/drivers/uart/uart.c
```diff
@@ -1 +1,89 @@
...
+uart_handle_t uart_channel_map[UART_CHANNEL_COUNT];
+
+// Private ----------------------------
+
+/**
+ * The ISR triggered when the `huart` channel has completed a transmission
+ */
+void uart_transmit_complete_isr(UART_HandleTypeDef *huart)
+{
+    // TODO: do stuff...
```
celery6, line 31: image
this function is called after the hal_uart_transmit_it() is called from uart_read

src/drivers/uart/uart.c
```diff
@@ -1 +1,89 @@
...
+    }
+    if (xSemaphoreTake(uart_channel_map[channel].write_mutex, timeout) != pdTRUE)
+    {
+        return W_IO_TIMEOUT; // Could not acquire the mutex in the given time
+    }
+    HAL_StatusTypeDef transmit_status = HAL_UART_TRANSMIT_IT(uart_channel_map[channel], data, len);
+    if (transmit_status == HAL_ERROR)
+        status = W_IO_ERROR;
+    else if (transmit_status == HAL_TIMEOUT)
+        status = W_IO_TIMEOUT;
```
martendo, line 75:
```suggestion
    if (HAL_ERROR == transmit_status) {
        status = W_IO_ERROR;
    } else if (HAL_TIMEOUT == transmit_status) {
        status = W_IO_TIMEOUT;
    }
```
as per BARR-C:2018 1.3a (braces) and 8.6a (constant on left)

celery6 (changes requested):
this looks all good so far, there's just more to be added (and it's all detailed in the design doc conveniently!)

src/drivers/uart/uart.c
```diff
@@ -1 +1,89 @@
...
+}
+
+// Get latest full msg received by `channel`
+// Will drop msgs if msgs are received faster than they are read
+// Blocks for up to `timeout` ms if no msg is available
+// Returns the status, received msg into `data`, msg length into `len`
+w_status_t uart_read(uart_channel_t channel, uint8_t *data, uint8_t *len, uint32_t timeout)
+{
+    w_status_t status = W_SUCCESS;
+    // aaaaaaaaaaaa
```
martendo, line 87:
```suggestion
    // TODO
```
minor, but being able to search the repo for the string `TODO` can be helpful (BARR-C:2018 2.2i.iii)

martendo (changes requested):
This PR also needs reformatting, try `scripts/format.sh`

src/drivers/uart/uart.c
```diff
@@ -1 +1,101 @@
...
+    UART_HandleTypeDef *handle; // HAL uart handle
+    SemaphoreHandle_t
+        transfer_complete;         // Communicate transfer complete between uart_write and the ISR
+    SemaphoreHandle_t write_mutex; // Allows tasks to line up to use uart_write
+} uart_handle_t;
+
+/**
+ * Maps each `uart_channel_t` to its `uart_handle_t` data
+ */
+uart_handle_t uart_channel_map[UART_CHANNEL_COUNT];
```
martendo, line 22: Do we ever initialize this array? Consider that `uart_init()` expects the array to already be ready for writing.

martendo (changes requested):

martendo:
Are we forgetting to include `stm32h7xx_hal_uart.h`? Looks like build fails with a lot of missing function declaration warnings, but I don't have the knowledge to be sure this is what we're missing.

src/drivers/uart/uart.c
martendo, line 22:
```suggestion
static uart_handle_t uart_channel_map[UART_CHANNEL_COUNT] = {0};
```

src/drivers/uart/uart.c
```diff
@@ -1 +1,101 @@
...
+    return status;
+}
+
+// Write to the specified UART channel
+// One task can write to a channel at once, and concurrent calls will block for `timeout`
+// Returns the status of data transfer
+w_status_t uart_write(uart_channel_t channel, const uint8_t *data, uint8_t len, uint32_t timeout)
+{
+    w_status_t status = W_SUCCESS;
+    if (UART_CHANNEL_COUNT <= channel || uart_channel_map[channel].handle == NULL)
```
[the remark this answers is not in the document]
samcyx: Ah okay thanks for the clarification!!

src/drivers/uart/uart.c
```diff
@@ -83,13 +97,57 @@ w_status_t uart_init(uart_channel_t channel, UART_HandleTypeDef *huart, uint32_t
...
+ * @brief Write to the specified UART channel
+ * @details // One task can write to a channel at once, and concurrent calls will block for 'timeout
+ * @param channel UART channel to write to
+ * @param data Buffer to store the data to send
+ * @param length Pointer to store the length of the sending message
+ * @param timeout_ms Timeout in milliseconds
+ * @return Status of the write operation
+ */
+w_status_t
+uart_write(uart_channel_t channel, uint8_t *buffer, uint8_t *length, uint32_t timeout_ms) {
```
martendo, line 124: I'm a little confused about the type of `length` here: (1) if `UART_MAX_LEN` is 256 it won't fit in 8 bits, (2) the design doc says `length` would be `uint16_t *`, also (3) the corresponding parameter to `HAL_UART_Transmit_IT` is `uint16_t`.

On that note, is there a reason why `uart_msg_t`'s `len` field is 32-bit if these functions and the HAL functions only accept a 16-bit size?

src/drivers/uart/uart.c
```diff
@@ -83,13 +97,57 @@ w_status_t uart_init(uart_channel_t channel, UART_HandleTypeDef *huart, uint32_t
...
+        return W_IO_TIMEOUT; // Could not acquire the mutex in the given time
+    }
+    HAL_StatusTypeDef transmit_status =
+        HAL_UART_Transmit_IT(s_uart_handles[channel].huart, buffer, *length);
+
+    if (HAL_ERROR == transmit_status) {
+        status = W_IO_ERROR;
+    } else if (HAL_TIMEOUT == transmit_status) {
+        status = W_IO_TIMEOUT;
+    }
```
martendo, line 139: This isn't in the API description and I didn't check last time but looking at the [source code](https://github.com/waterloo-rocketry/cansw_processor_canards/blob/dad939f2d5f92fd0b0d819f657f68dedd50e3ab2/Drivers/STM32H7xx_HAL_Driver/Src/stm32h7xx_hal_uart.c#L1281-L1347) apparently `HAL_UART_Transmit_IT` will only ever return `HAL_OK`, `HAL_ERROR`, or `HAL_BUSY`, so that last condition would never be true. Should we be checking for `HAL_BUSY` instead of `HAL_TIMEOUT`?

tests/unit/uart_test.cpp
```diff
@@ -107,13 +120,25 @@ TEST_F(UartTest, InitFailures) {
...
+TEST_F(UartTest, WriteSuccess) {
+    uint8_t buffer[UART_MAX_LEN];
+    uint8_t length;
+    xSemaphoreTake_fake.return_val = pdTRUE;
+    xSemaphoreGive_fake.return_val = pdTRUE;
+    w_status_t status = uart_write(TEST_CHANNEL, buffer, &length, timeout);
+    EXPECT_EQ(W_SUCCESS, status);
+}
+
+// TODO: add more tests for uart_write
```
martendo, line 140: Some good tests: invalid parameters, semaphore functions time out (return `pdFALSE`), HAL function returns bad status

tests/unit/uart_test.cpp
```diff
@@ -107,13 +120,25 @@ TEST_F(UartTest, InitFailures) {
...
 }
 
+// Test uart_write with valid message
+TEST_F(UartTest, WriteSuccess) {
+    uint8_t buffer[UART_MAX_LEN];
+    uint8_t length;
+    xSemaphoreTake_fake.return_val = pdTRUE;
+    xSemaphoreGive_fake.return_val = pdTRUE;
+    w_status_t status = uart_write(TEST_CHANNEL, buffer, &length, timeout);
+    EXPECT_EQ(W_SUCCESS, status);
```
martendo, line 137: Maybe we could also check that the mutex and semaphore behaviour is correct?
```suggestion
    EXPECT_EQ(W_SUCCESS, status);
    EXPECT_EQ(2, xSemaphoreTake_fake.call_count);
    EXPECT_EQ(1, xSemaphoreGive_fake.call_count);
```

martendo (dismissed):
I think I went a little too in-depth with this review but hey it looks pretty cool! Almost there :muscle:

src/drivers/uart/uart.c
```diff
@@ -83,13 +97,57 @@ w_status_t uart_init(uart_channel_t channel, UART_HandleTypeDef *huart, uint32_t
...
+ * @param timeout_ms Timeout in milliseconds
+ * @return Status of the write operation
+ */
+w_status_t
+uart_write(uart_channel_t channel, uint8_t *buffer, uint8_t *length, uint32_t timeout_ms) {
+    w_status_t status = W_SUCCESS;
+    if ((channel >= UART_CHANNEL_COUNT) || (NULL == s_uart_handles[channel].huart)) {
+        status = W_INVALID_PARAM; // Invalid parameter(s)
+        return status;
+    } else if (pdTRUE != xSemaphoreTake(s_uart_handles[channel].write_mutex, timeout_ms)) {
```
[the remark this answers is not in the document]
celery6: implicitly converting between ms and ticks was actually a source of confusion last year cuz its not obvious or even guaranteed that 1tick=1ms (what if i changed the tick rate on a whim?). being explicit like this is so much nicer

Joe-Joe-Joe-Joe (approved):
LGTM

Joe-Joe-Joe-Joe:
Not clear where along here this was tested IRL. I will add to my list of stuff to test tomorrow.

celery6:
@Joe-Joe-Joe-Joe this was tested irl during those commits 5 days ago. i was literally there too. doesnt look like any source code changed since then so idt u need to test again

celery6 (approved):

PR #108, opened 2025-02-06
Specify platform for Docker image compatibility

taisirhassan:
Ensure the Docker image is compatible with the arm-none-eabi gcc toolchain by explicitly specifying the platform of the image (if left unspecified, docker downloads the image based on the user's architecture system, big issue for mac users if left unspecified because of build failures). potential investigatory options are as follows regarding this: 

we can either a) hate on mac users on arm64 by forcing them to run x86 images and the binaries through rosetta 2 (tbh not even that slow) or b) change our dockerfile to download a diff toolchain architecture based on the host architecture kind of a hassle but not too annoying of a change. this pr implements option a).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,5 +1,5 @@
 # Use Ubuntu as the base image
-FROM ubuntu:22.04
+FROM --platform=linux/amd64 ubuntu:22.04
 
 # Set non-interactive mode for apt-get
 ENV DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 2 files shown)


celery6 (approved):
WHY MAC WHYYYYYYYYYYYYYYYYYYYYY

celery6:
admin abuse merging because this fixes extrmeely broken mac :skullemoji

SAT-oO:
if cmake doesn't like you when you run the script maybe try pressing on the gear button in the bottom blue bar or typing in the command to let it build 💀

PR #111, opened 2025-02-06
Implement UART read functionality

taisirhassan:
This PR implements UART read support for the processor board, enabling reliable data reception from the Movella IMU and debug serial.
Key features:

Interrupt-driven UART reception with FreeRTOS synchronization
Message completion detection via IDLE frame or buffer full
Timeout handling and error checking
Parameter validation


src/drivers/uart/uart.c
```diff
@@ -1 +1,196 @@
...
+        handle->rx_msgs[i].len = 0;
+        handle->rx_msgs[i].busy = false;
+    }
+
+    /* Start with buffer 0 for initial reception */
+    handle->curr_write_ind = 0;
+
+    /* Create queue for passing complete messages to readers.
+     * Single slot queue - we only keep most recent message. */
+    handle->avail_msgs = xQueueCreate(1, sizeof(uart_msg_t));
```
celery6, line 56: i realize the data arr makes uart_msg_t super fat and thick which is not optimal for freertos queue which passes by copy. is it possible to make this queue uart_msg_t*, or make the data arr in uart_msg_t to be a pointer to some statically allocated thing? the point is, avoid passing around the large object by copy

src/drivers/uart/uart.c
```diff
@@ -1 +1,196 @@
...
+    *length = msg_len;
+
+    return W_SUCCESS;
+}
+
+/* HAL callback for UART IDLE line detection.
+ * Called from interrupt context when message completes. */
+void HAL_UARTEx_RxEventCallback(UART_HandleTypeDef *huart, uint16_t size) {
+    /* Delegate to handler to keep ISR simple */
+    handle_idle_event(huart, size);
```
celery6, line 105: why does this funny wrapping exist, just use hal registercallback thing which is proect-wide convention

src/drivers/uart/uart.c
```diff
@@ -1 +1,196 @@
...
+        if (s_uart_handles[ch].huart == huart) {
+            break;
+        }
+    }
+    if (ch >= UART_CHANNEL_COUNT) {
+        return; /* Not one of our UARTs */
+    }
+
+    uart_handle_t *handle = &s_uart_handles[ch];
+    uint8_t w_ind = handle->curr_write_ind;
```
celery6, line 125: request to rename these `ind` to smth like `buffer_num` or smth cuz i started thinking this is a byte write index help

src/drivers/uart/uart.c
```diff
@@ -1 +1,196 @@
...
+    msg->len = size;
+    msg->busy = true;
+
+    /* Make message available to readers. Uses overwrite version
+     * so newest message is never blocked by old unread message. */
+    xQueueOverwriteFromISR(handle->avail_msgs, msg, &higher_priority_task_woken);
+
+    /* Try to switch to other buffer for next reception.
+     * Only switch if other buffer isn't still busy from last time. */
+    uint8_t next_ind = (w_ind + 1U) % 2U;
```
celery6, line 138: can 2 be a macro define "num_buffers" or smth

celery6: this concept theoretically could be expanded to more buffers .. right?

src/drivers/uart/uart.c
celery6, line 1: ur end of files dont have newline :sob:

src/drivers/uart/uart.c
```diff
@@ -1 +1,196 @@
...
+    /* Wait for message to become available.
+     * xQueueReceive blocks until message arrives or timeout. */
+    if (xQueueReceive(handle->avail_msgs, &msg, pdMS_TO_TICKS(timeout_ms)) != pdTRUE) {
+        *length = 0;
+        return W_IO_TIMEOUT;
+    }
+
+    /* Copy message to user's buffer. Length check prevents buffer overflow,
+     * though MAX_MSG_LEN should already enforce this. */
+    uint16_t msg_len = (uint16_t)msg.len;
```
celery6, line 91: why casting this seems very sus
if the isr returns a size >16bits and u cast it then.. huh? does it make more sense to ust use uint32t for this field and enforce size by checking it instead of implication by uint type

src/drivers/uart/uart.c
```diff
@@ -1 +1,196 @@
...
+    if (msg_len > MAX_MSG_LEN) {
+        msg_len = MAX_MSG_LEN;
```
celery6, line 93: add a TODO for recording overflows like this

celery6 (changes requested):
functionality looks good now i think this would work in 15mins!!

src/drivers/uart/uart.c
taisirhassan: yes it can cuz modulo arithmetic lol!

celery6 (approved):
LFGTMTFTHFHGMH!!!!!!! hopeuflly it works with movella :face

PR #112, opened 2025-02-07
Can handler

JacobG1003:
Initial Implementation


src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
+#include "can_handler.h"
+
+QueueHandle_t busQueue_rx;
+QueueHandle_t busQueue_tx;
```
martendo, line 4:
```suggestion
QueueHandle_t busQueue_rx = NULL;
QueueHandle_t busQueue_tx = NULL;
```
as per BARR-C:2018 7.2d: "Any pointer variable lacking an initial address shall be initialized to `NULL`."

Also, these variable names aren't exactly `snake_case`.

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+
+can_callback_t callback_map[30];
```
martendo, line 6:
```suggestion
static can_callback_t callback_map[30] = {0};
```
use `static` as per BARR-C:2018 1.8a, and initialize to zeros

Initializing this to zeros is important!!! The code below assumes that an unregistered callback pointer will have the value of `NULL`, but without initialization this is not defined behaviour!

martendo, line 6: I wonder if we should add a value to `can_msg_type_t` with the max value of any message type so we don't have to hard-code 30 as the array size here.

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+
+w_status_t can_register_callback(can_msg_type_t msg_type, can_callback_t callback)
+{
+    callback_map[msg_type] = callback;
+    return W_SUCCESS;
+}
+
+void can_handler_tx(const can_msg_t *message)
+{
+    if (xQueueSend(busQueue_tx, message, 10))
```
martendo, line 24:
```suggestion
    if (xQueueSend(busQueue_tx, message, 10) == pdTRUE)
```

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+{
+    for (;;)
+    {
+        can_msg_t tx_msg;
+        // Block the thread until we see data in the bus queue or 1 sec elapses
+        if (xQueueReceive(busQueue_tx, &tx_msg, 1000) == pdTRUE)
+        { // Returns pdTRUE if we got a message, pdFALSE if timed out
+            if (can_send(&tx_msg) == false)
+            {
+                logError("CAN", "CAN send failed!");
```
martendo, line 60:
```suggestion
                logText("CAN", "CAN send failed!");
```
the new function is called `logText`

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+{
+    busQueue_rx = xQueueCreate(16, sizeof(can_msg_t));
+    busQueue_tx = xQueueCreate(16, sizeof(can_msg_t));
+
+    if (busQueue_tx == NULL)
+        return W_FAILURE;
+    if (busQueue_rx == NULL)
+        return W_FAILURE;
+    if (!can_init_stm(&hfdcan1, can_handle_rx_isr))
+        return W_FAILURE;
```
martendo, line 78:
```suggestion
    if (NULL == busQueue_tx) {
        return W_FAILURE;
    } else if (NULL == busQueue_rx) {
        return W_FAILURE;
    } else if (!can_init_stm(&hfdcan1, can_handle_rx_isr)) {
        return W_FAILURE;
    }
```
as per BARR-C:2018 8.6a and 1.3a

martendo (dismissed):
This PR also needs reformatting, try `scripts/format.sh`

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+        can_msg_t tx_msg;
+        // Block the thread until we see data in the bus queue or 1 sec elapses
+        if (xQueueReceive(busQueue_tx, &tx_msg, 1000) == pdTRUE)
+        { // Returns pdTRUE if we got a message, pdFALSE if timed out
+            if (can_send(&tx_msg) == false)
+            {
+                logError("CAN", "CAN send failed!");
+            }
+            vTaskDelay(1);
+            HAL_GPIO_TogglePin(GPIOE, GPIO_PIN_10); // write and toggle D3 when we send a CAN message
```
Joe-Joe-Joe-Joe, line 63: I believe we have a driver for this now?

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+w_status_t can_handler_init(void)
+{
+    busQueue_rx = xQueueCreate(16, sizeof(can_msg_t));
+    busQueue_tx = xQueueCreate(16, sizeof(can_msg_t));
+
+    if (busQueue_tx == NULL)
+        return W_FAILURE;
+    if (busQueue_rx == NULL)
+        return W_FAILURE;
+    if (!can_init_stm(&hfdcan1, can_handle_rx_isr))
```
Joe-Joe-Joe-Joe, line 77: This should be changed from extern-ing the handle to passing the pointer to the handle as a parameter to `can_handler_init`

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+    {
+        return W_FAILURE;
+    }
+}
+
+void can_handler_task_rx(void *argument)
+{
+    for (;;)
+    {
+        can_msg_t *rx_msg;
```
Joe-Joe-Joe-Joe, line 38:
```suggestion
        can_msg_t rx_msg;
```

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+        if (xQueueReceive(busQueue_rx, &rx_msg, 10) == pdTRUE)
+        {
+            uint16_t msgType = get_message_type(rx_msg);
```
Joe-Joe-Joe-Joe, line 41:
```suggestion
            uint16_t msgType = get_message_type(&rx_msg);
```

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,80 @@
...
+            if (callback_map[msgType] != NULL)
+            {
+                callback_map[msgType](rx_msg);
```
Joe-Joe-Joe-Joe, line 44:
```suggestion
                callback_map[msgType](&rx_msg);
```

src/application/can_handler/can_handler.h
```diff
@@ -11,6 +11,8 @@
 
 #include "queue.h"
```
Joe-Joe-Joe-Joe, line 12: Does this/Freertos.h still need to be in the header?

Joe-Joe-Joe-Joe (changes requested):

src/application/can_handler/can_handler.c
celery6: oh i wanted to do that too 
https://github.com/waterloo-rocketry/canlib feel free to open a pr jason will probably approve it!

src/application/can_handler/can_handler.c
celery6: we so do! and i think we can get rid of this behaviour this year? why did it exist in production last year? i think this time we might use leds as a last resort error reporting,so dont use it here

celery6 (commented):
looks very cool
address minor changes and test rigorously

src/application/can_handler/can_handler.c
Joe-Joe-Joe-Joe: Wouldn't it be more clear to initialize to `NULL`?

martendo: As in `= {NULL}`? Yeah, I agree &mdash; updated the suggestion

src/application/can_handler/can_handler.c
```diff
@@ -0,0 +1,98 @@
...
+    if (busQueue_tx == NULL) {
+        return W_FAILURE;
+    }
+    if (busQueue_rx == NULL) {
+        return W_FAILURE;
+    }
+    if (!can_init_stm(&hfdcan1, can_handle_rx_isr)) {
+        return W_FAILURE;
+    }
+    return W_SUCCESS;
```
Joe-Joe-Joe-Joe, line 73: We should also bind the callback for reset commands `MSG_RESET_CMD` here, since there isn't a better place to put it and we want to be able to reset so long as CAN is working

Joe-Joe-Joe-Joe (changes requested):
Please get more eyes on https://github.com/waterloo-rocketry/canlib/pull/71 and approve first

Joe-Joe-Joe-Joe:
Blocked on 
https://github.com/waterloo-rocketry/canlib/pull/74 
https://github.com/waterloo-rocketry/canlib/pull/75 
Should be merged by EOD

celery6:
closes https://github.com/waterloo-rocketry/cansw_processor_canards/issues/94 ?

src/application/can_handler/can_handler.c
```diff
@@ -1,8 +1,114 @@
...
+static QueueHandle_t bus_queue_tx = NULL;
+static uint32_t dropped_rx_counter = 0;
+
+static can_callback_t callback_map[30] = {NULL};
+
+static w_status_t can_reset_callback(const can_msg_t *msg) {
+    if (check_board_need_reset(msg)) {
+        NVIC_SystemReset();
+    }
+    return W_SUCCESS;
```
celery6, line 16: did u test the reset to work?
Also if this function returns after it's supposed to reset, it should be a failure :face:

src/application/can_handler/can_handler.c
```diff
@@ -1,8 +1,114 @@
...
+}
+
+static w_status_t can_led_on_callback(const can_msg_t *msg) {
+    (void)msg;
+    gpio_write(GPIO_PIN_RED_LED, 0, 5);
+    gpio_write(GPIO_PIN_GREEN_LED, 0, 5);
+    gpio_write(GPIO_PIN_BLUE_LED, 0, 5);
```
celery6, line 23: use the level enum for the 2nd param instead of rawdogging 0/1... ? nit

celery6: gpio_write returns a status thats not guaranteed to be SUCESS. u have to check it. ik this is trivial case. nit moment

src/application/can_handler/can_handler.c
```diff
@@ -1,8 +1,114 @@
...
+    return W_SUCCESS;
+}
+
+w_status_t can_handler_register_callback(can_msg_type_t msg_type, can_callback_t callback) {
+    callback_map[msg_type] = callback;
     return W_SUCCESS;
 }
+
+w_status_t can_handler_transmit(const can_msg_t *message) {
+    if (pdPASS == xQueueSend(bus_queue_tx, message, 10)) {
```
celery6, line 79: should we really let this block for up to 10ms? that would break the control loop.

.. adding to TODO list to look into all the arbitrary timeouts in the codebase :dies`

src/application/can_handler/can_handler.c
```diff
@@ -1,8 +1,114 @@
...
+        }
+    }
+}
+
+void can_handler_task_tx(void *argument) {
+    (void)argument;
+    for (;;) {
+        can_msg_t tx_msg;
+        // Block the thread until we see data in the bus queue or 1 sec elapses
+        if (pdPASS == xQueueReceive(bus_queue_tx, &tx_msg, 100)) {
```
celery6, line 105: 100ms is not 1sec as the comment suggests

celery6 (approved):
some assorted nits that arent breaking anything so LGTM!!!!!!!!!

src/application/can_handler/can_handler.c
Joe-Joe-Joe-Joe: I forgor :face:

Are you suggesting I add a return `W_FAILURE` inside of the `if`? if the system reset fails we have other issues. If we get a reset message that _isn't_ for us, then a success is the correct return code (we successfully parsed and handled the message)

src/application/can_handler/can_handler.c
Joe-Joe-Joe-Joe: Forgot this is a real queue, you are correct

Joe-Joe-Joe-Joe:
Tested reset handler IRL
- Resets board when `board_type_id` is ANY or when `board_type_id` is PROCESSOR and `board_inst_id` is GENERIC
- Does not reset the board when it shouldn't (not tested exhaustively, just a couple cases)

Rx and Tx still work as expected

BEWARE OF FAULTY USBDEBUG FML

celery6 (approved):

PR #116, opened 2025-02-08
Controller implementation: freertos

SAT-oO:


src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
...
+// internal: flight_condition_t
+typedef struct {
+    float dynamic_pressure;
+    float canard_coeff_lift;
+} flight_condition_t;
+
+flight_condition_t flight_condition;
+controller_t controller_state;
+controller_input_t controller_input;
+controller_output_t estimator_controller_input;
```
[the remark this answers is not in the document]
celery6: this comment applies to all the variables in this file that u want to make "private"

src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
...
+
+/*
+Send `canard_angle`, the desired canard angle (radians) to CAN
+*/
+w_status_t controller_send_san(float canard_angle) {
```
celery6, line 23: typo

src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
...
+    return W_SUCCESS;
+}
+
+/**
+ * Initialize controller module
+ * Must be called before RTOS scheduler starts
+ * @return W_SUCCESS if initialization successful
+ */
+w_status_t controller_init(void) {
+    bool init_status = true;
```
celery6, line 35: why not just make init_status of type w_status_t to start with? that's the whole point of it. see other modules for example

src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
 #include "application/controller/controller.h"
 
+QueueHandle_t inputQueue;
+QueueHandle_t internalStateQueue;
+QueueHandle_t outputQueue;
```
celery6, line 5: where snack_case

src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
...
+    inputQueue = xQueueCreate(1, sizeof(controller_input_t)); // updated by estimator
+    internalStateQueue =
+        xQueueCreate(1, sizeof(controller_input_t)); // estimator function to controller task
+
+    // check queue creation
+    if (internalStateQueue == NULL) {
+        init_status = false;
+    }
+
+    // initialize structs: placeholder values
```
celery6, line 47: i think u should initialize the structs to {0} at the lines they're declared.
like `flight_condition_t  flight_condition= {0}`.
Then only set non-zero values here.

Or if the initialization values are gonna be constant, set the initial values at declaration instead of in this function?
`flight_condition_t flight_condition = { .canard_coeff_lift = 99,  .dynamic_pressure = 1}`

src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
...
+    // return w_status_t state
+    return W_SUCCESS;
+}
+
+/**
+ * Update controller with new state data - called by state estimation module
+ * @param new_state Latest state estimate from state estimation
+ * @return W_FAILURE if validation/queueing fails
+ */
+w_status_t controller_update_inputs(controller_input_t *new_state) {
```
celery6, line 106: i think u misunderstood how queues work for this and the get latest output function

src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
...
+        "controller", "controller output didn't send, command angle %f", output->commanded_angle
+    );
+    return W_FAILURE;
+}
+
+/**
+ * Controller task function for RTOS
+ */
+void controller_task(void *argument) {
+    // get current flight phase
```
celery6, line 149: remember freertos tasks have to run an infinite loop
see examples

src/application/controller/controller.c
```diff
@@ -1,2 +1,185 @@
...
+    flight_phase_state_t current_phase = STATE_INIT;
+
+    // log phase transitions, specifics logged in flight phase
+    if (current_phase != flight_phase_get_state()) {
+        logInfo("controller", "flight phase changed");
+    }
+
+    if (current_phase != STATE_ACT_ALLOWED || current_phase != STATE_COAST) { // if not proper state
+        vTaskDelay(pdMS_TO_TICKS(1)); // delay 1 ms
```
celery6, line 158: if not the proper state, the rest of this function must not run. rn u make it delay for 1ms, then run the rest of it anyway

celery6 (changes requested):

src/application/controller/controller.c
```diff
@@ -1,2 +1,140 @@
...
+    return init_status;
+}
+
+/**
+ * Update controller with new state data - called by state estimation module
+ * @param new_state Latest state estimate from state estimation
+ * @return W_FAILURE if validation/queueing fails
+ */
+w_status_t controller_update_inputs(controller_input_t *new_state) {
+    if (xQueueOverwrite(internal_state_queue, new_state) == pdPASS) { // overwrite internal queue
```
Joe-Joe-Joe-Joe, line 70: `xQueueOverwrite` can only ever return `pdPASS` so this error checking is ineffective
https://www.freertos.org/Documentation/02-Kernel/04-API-references/06-Queues/11-xQueueOverwrite

src/application/controller/controller.c
```diff
@@ -1,2 +1,140 @@
...
+/**
+ * Controller task function for RTOS
+ */
+void controller_task(void *argument) {
+    // get current flight phase
+    flight_phase_state_t current_phase = STATE_INIT;
+
+    while (true) {
+        // log phase transitions, specifics logged in flight phase
+        if (current_phase != flight_phase_get_state()) {
```
Joe-Joe-Joe-Joe, line 104: You also need to update the current phase

src/application/controller/controller.c
```diff
@@ -1,2 +1,140 @@
...
+            log_text("controller", "new state data received for controller computations");
+
+        } else {
+            log_text("controller", "no new state data received for controller computations");
+            controller_state.data_miss_counter++;
+
+            // if number of data misses exceed threshold, transition to safe mode
+        }
+
+        // calculate gains based on current conditions
```
Joe-Joe-Joe-Joe, line 126: I am confused how this work is being divided, cause clearly most of the implementation (translation from MATLAB) is not complete here

Joe-Joe-Joe-Joe (changes requested):

src/application/controller/controller.c
SAT-oO: 💀 skill issue

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+static controller_t controller_state = {0};
+static controller_input_t controller_input = {0};
+static controller_output_t controller_output = {0};
+static gain_table_entry_t gain_table_entry = {0};
+
+/*
+    TODO Send `canard_angle`, the desired canard angle (radians) to CAN
+*/
+static w_status_t controller_send_can(float canard_angle) {
+    // Build the CAN msg using [canard-specific canlib function to be defined
```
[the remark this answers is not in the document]
Joe-Joe-Joe-Joe: You probably want
- `PRIO_HIGH`
- `ACTUATOR_CANARD_ANGLE`
- `timer_get_ms` for timestamp

and then call `can_handle_tx` on your can message struct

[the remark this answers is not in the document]
Joe-Joe-Joe-Joe: @rexblade21 how does `ACTUATOR_CANARD_ENABLE` work? Does proc also need to send this?

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+    output_queue = xQueueCreate(1, sizeof(controller_output_t));
+
+    // check queue creation
+    if (internal_state_queue == NULL || output_queue == NULL) {
+        init_status = W_FAILURE;
+        log_text("controller", "queue creation failed");
+    }
+
+    // gain table file memory test
+    gain_table = fopen("file path", "r");
```
Joe-Joe-Joe-Joe, line 54: huh? There is no standard file i/o cause there is no filesystem. I think this is irrelevant either way cause we are going to put the table in code? Anyway this section should be deleted

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+
+    if (gain_table == NULL) {
+        log_text("controller", "gain table invalid");
+        init_status = W_FAILURE;
+    } else {
+        log_text("controller", "gain table file opened");
+    }
+
+    // Log initialization status
+    if (init_status == W_SUCCESS) {
```
Joe-Joe-Joe-Joe, line 64: I think this is borderline too much logging. We will know the init status based on the return code so i think this is redundant.

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+
+/**
+ * Get most recent control output (cammanded angle and timestamp) - called by
+ * state estimation module
+ * @param output Pointer to store output -> type defined in controller
+ * @return W_FAILURE if no output available
+ */
+w_status_t controller_get_latest_output(controller_output_t *output) {
+    if (xQueuePeek(output_queue, output, timeout) == pdPASS) {
+        log_text("controller", "latest output returned");
```
Joe-Joe-Joe-Joe, line 94: Again, redundant logging. In general, if something has to run every 5ms like controller/SE, we should never log nominal conditions

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+    log_text("controller", "timeout: no output available");
+    return W_FAILURE;
+}
+
+/**
+ * Controller task function for RTOS
+ */
+void controller_task(void *argument) {
+    // get current flight phase
+    flight_phase_state_t current_phase = STATE_INIT;
```
Joe-Joe-Joe-Joe, line 106: fwiw flight_phase_get_state() is guaranteed to return a state (the internal queue is populated at initialization) after scheduler start, so you can call it here. That is more robust in case we change the starting phase (for whatever reason)

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+w_status_t controller_init(void) {
+    w_status_t init_status = W_SUCCESS;
+
+    // Create internal state queue (length = 1)
+    internal_state_queue =
+        xQueueCreate(1, sizeof(controller_input_t)); // estimator function to controller task
+    output_queue = xQueueCreate(1, sizeof(controller_output_t));
+
+    // check queue creation
+    if (internal_state_queue == NULL || output_queue == NULL) {
```
Joe-Joe-Joe-Joe, line 48:
```suggestion
    if ((NULL == internal_state_queue) || (NULL == output_queue)) {
```
something something BARR standard i dont remember which rule this is, i just know you are supposed to do constant == variable.

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+
+    while (true) {
+        // log phase transitions, specifics logged in flight phase
+        if (current_phase != flight_phase_get_state()) {
+            current_phase = flight_phase_get_state();
+            log_text("controller", "flight phase changed");
+        }
+
+        if (current_phase != STATE_ACT_ALLOWED &&
+            current_phase != STATE_COAST) { // if not in proper state
```
Joe-Joe-Joe-Joe, line 116: `STATE_COAST` is also fake. `STATE_ACT_ALLOWED ` is the only valid state for controller actuation. That one is on me oops

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+            vTaskDelay(pdMS_TO_TICKS(1));
+        } else {
+            // wait for new state data (5ms timeout)
+            controller_input_t new_state_msg;
+            if (xQueueReceive(internal_state_queue, &new_state_msg, timeout) == pdPASS) {
+                controller_state.current_state = new_state_msg;
+                // TODO validate data
+
+                // log data received
+                log_text("controller", "new state data received for controller computations");
```
Joe-Joe-Joe-Joe, line 126: Don't log success on periodic operations

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+            } else {
+                log_text("controller", "no new state data received for controller computations");
+                controller_state.data_miss_counter++;
+
+                // TODO if number of data misses exceed threshold, transition to safe mode
+            }
+
+            // calculate gains based on current conditions
+            uint32_t timestamp = controller_state.current_state.timestamp;
+            if(timestamp > 5){
```
Joe-Joe-Joe-Joe, line 137: I saw finn mention this somewhere else but this is a really opaque way to set the reference. we should find a clearer way. In particular `gain_table_entry.reference_signal` is difficult to interpret, this really should be the roll rate (is this the roll rate???)

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+                }else if(timestamp > 26){
+                    gain_table_entry.reference_signal = 0;
+                }
+            }
+
+            // TODO Ks: load table, interpolate
+            // deconstruct Ks into K and K_pre
+
+            // compute control output and overwrites output queue
+            xQueueOverWrite(output_queue, &controller_output);
```
Joe-Joe-Joe-Joe, line 151: Capitalization? Does this even compile

src/application/controller/controller.c
```diff
@@ -1,2 +1,166 @@
...
+            // send command visa CAN + log status/errors
+            if(controller_send_can(controller_output.commanded_angle) == W_SUCCESS){
+                log_text("controller", "commanded angle sent via CAN");
+
+            }else{
+                log_text("controller", "commanded angle failed to send via CAN");
+            }
+
+
+            // TODO send feedback gain to estimator [gain_k]
```
Joe-Joe-Joe-Joe, line 162: isn't sending controller data to estimator already done via the queue above?

src/application/controller/controller.c
SAT-oO: I renamed it to be just reference_signal in the controller-calc branch but the entire roll program shouldn't be here for test flight. will stay removed for now

src/application/controller/controller.c
```diff
@@ -1,2 +1,124 @@
 #include "application/controller/controller.h"
+#include "application/controller/gain_scheduling.h"
+#include "application/flight_phase/flight_phase.h"
+#include "application/logger/log.h"
+#include "queue.h"
+
+static QueueHandle_t internal_state_queue;
+QueueHandle_t output_queue;
```
Joe-Joe-Joe-Joe, line 8: this should also be static?

src/application/controller/controller.c
```diff
@@ -1,2 +1,124 @@
...
+
+static const TickType_t timeout = pdMS_TO_TICKS(5);
```
Joe-Joe-Joe-Joe, line 10: Thinking more about this, we should adopt a convention for freertos timeouts since we use them everywhere. I've be `#define`ing a timeout in ms that is as verbosely named as possible, then leaving the call to `pdMS_TO_TICKS` inside each freertos function call. This comes up in stuff like UART or ADC where we might have both a mutex to use a driver and an internal semaphore and they both have their own timeouts (which add, from the perspective of the calling function).

@celery6 thoughts? 

Either way I think `timeout` is too generic, should specify in name or in comment what timeout this represents

src/application/controller/controller.c
```diff
@@ -1,2 +1,124 @@
...
+ */
+void controller_task(void *argument) {
+    (void)argument;
+    // get current flight phase
+    flight_phase_state_t current_phase = flight_phase_get_state();
+
+    while (true) {
+        // log phase transitions, specifics logged in flight phase
+        if (current_phase != flight_phase_get_state()) {
+            current_phase = flight_phase_get_state();
```
Joe-Joe-Joe-Joe, line 88: I think this is technically a race condition? The return value of `flight_phase_get_state` _can_ change between these two calls. You should store the state in a variable, then do the comparison and assignment with that

src/application/controller/controller.c
```diff
@@ -1,2 +1,124 @@
...
+
+                // no roll program for test flight
+
+            } else {
+                controller_state.data_miss_counter++;
+
+                // TODO if number of data misses exceed threshold, notify health check module
+            }
+
+            // TODO controller calc: interpolate
```
Joe-Joe-Joe-Joe, line 109: @FinnBreu can you clarify intended behavior? Do we run the controller even if the state data doesn't get updated? I would think if data takes like, 6ms, we would want to wait for 5ms, log a data miss, and then try again, and then successfully get the state data (after another 1ms of waiting)?

Joe-Joe-Joe-Joe: i.e. should this be inside the `if` above?

src/application/controller/controller.c
```diff
@@ -1,2 +1,124 @@
...
+
+            // update output queue
+            xQueueOverwrite(output_queue, &controller_output);
+
+            // send command visa CAN + log status/errors
+            if (controller_send_can(controller_output.commanded_angle) == W_SUCCESS) {
+                log_text("controller", "commanded angle sent via CAN");
```
Joe-Joe-Joe-Joe, line 116: ahhhh success logging

Joe-Joe-Joe-Joe (changes requested):
Few minor things, looks mostly good

src/application/controller/controller.c
celery6: yeah timeouts should be defined in milliseconds and only translated to ticks at the last possible moment (ie when calling the freertos function).

src/application/controller/controller.c
```diff
@@ -1,2 +1,125 @@
...
+static controller_t controller_state = {0};
+static controller_input_t controller_input __attribute__((unused)) = {0};
+static controller_output_t controller_output = {0};
+static controller_gain_t controller_gain __attribute__((unused)) = {0};
+
+/*
+    TODO Send `canard_angle`, the desired canard angle (radians) to CAN
+*/
+static w_status_t controller_send_can(float canard_angle) {
+    // Build the CAN msg using [canard-specific canlib function to be defined later].
```
Joe-Joe-Joe-Joe, line 21: Please actually implement once #112 is merged

src/application/controller/controller.c
```diff
@@ -1,2 +1,125 @@
...
+}
+
+/**
+ * Get most recent control output (cammanded angle and timestamp) - called by
+ * state estimation module
+ * @param output Pointer to store output -> type defined in controller
+ * @return W_FAILURE if no output available
+ */
+w_status_t controller_get_latest_output(controller_output_t *output) {
+    if (xQueuePeek(output_queue, output, pdMS_TO_TICKS(CONTROLLER_CYCLE_TIMEOUT)) == pdPASS) {
```
Joe-Joe-Joe-Joe, line 70: invert constant and function in expression (BARR thing)

src/application/controller/controller.c
```diff
@@ -1,2 +1,125 @@
...
+ * Controller task function for RTOS
+ */
+void controller_task(void *argument) {
+    (void)argument;
+    // get current flight phase
+    flight_phase_state_t current_phase = flight_phase_get_state();
+    flight_phase_state_t prev_phase = current_phase;
+
+    while (true) {
+        // log phase transitions, specifics logged in flight phase
```
Joe-Joe-Joe-Joe, line 87:
```suggestion
        // log phase transitions, specifics logged in flight phase
        current_phase = flight_phase_get_state();
```
You need to check the current phase every cycle
if you don't do this, the `if` condition below will never be true

src/application/controller/gain_scheduling.h
```diff
@@ -0,0 +1,17 @@
+#ifndef GAIN_SCHEDULING_H_
+#define GAIN_SCHEDULING_H_
+
+#define GAIN_NUM 4
+#define FEEDBACK_GAIN_NUM 3
```
Joe-Joe-Joe-Joe, line 5:
```suggestion
#define FEEDBACK_GAIN_NUM (GAIN_NUM-1)
```
shouldn't have multiple independent lengths for defining a union

Joe-Joe-Joe-Joe (changes requested):
Just a few more nits, but looking much better

src/application/controller/controller.c
celery6: do that in another pr

celery6 (approved):
lgtm!!!!!!!!!!

src/application/controller/controller.c
```diff
@@ -1,2 +1,118 @@
...
+}
+
+/**
+ * Get most recent control output (cammanded angle and timestamp) - called by
+ * state estimation module
+ * @param output Pointer to store output -> type defined in controller
+ * @return W_FAILURE if no output available
+ */
+w_status_t controller_get_latest_output(controller_output_t *output) {
+    if (pdPASS == xQueuePeek(output_queue, output, pdMS_TO_TICKS(CONTROLLER_CYCLE_TIMEOUT_MS))) {
```
Joe-Joe-Joe-Joe, line 70: `CONTROLLER_CYCLE_TIMEOUT_MS` is called that cause it defines the cycle delay of the controller (i.e. the part in the task). Even if this timeout happens to be 5 ms, you should use a different macro because they represent unrelated delays

That being said, in this case the timeout should be 0, because the queue should always be populated. You should add a call to overwrite the output queue in `controller_init`, with just "0" or smth, otherwise you will be stuck in deadlock as soon as you initialize (controller won't run until it gets input from estimator, and estimator won't run until it gets input from controller)

Joe-Joe-Joe-Joe (changes requested):

celery6 (approved):
LGTMMMMMMMM

Issue #30, opened 2024-11-17
ST (LSM6DSV32X) imu module design


celery6:
see #5 
the interface for this should be as similar as possible to #12  and #20 to have consistent imu interfaces.
this module will be very similar to #12

celery6:
This imu is basically identical to the LSM6DO which is part of the polulu device. Only difference is it can measure g in range of +-32 which is more than the +-16 of the polulu.
datasheet and application note links: https://www.st.com/en/mems-and-sensors/lsm6dsv32x.html#documentation

PR #123, opened 2025-02-14
St imu

celery6:
yyu[pjppp

im pretty sure theres no other differences between this and the lsmd thing on the polulu right

#30


Joe-Joe-Joe-Joe (approved):
builds on my machine (wow)
cursory glance at the datasheet suggests this does in fact only do gyro/accel
LGTM

Redstone-ray (approved):

PR #144, opened 2025-02-25
IMU handler implementation

taisirhassan:
IMU Handler Implementation

This PR implements a multi-IMU data collection and processing system that interfaces with three IMU sensors:
- Polulu AltIMU-10 (accel, gyro, mag, baro)
- ST LSM6DSV32X (accel, gyro)
- Movella MTi-630 (accel, gyro, mag, baro)

Key features:
- 5ms sampling period for all sensors
- Robust error handling with data zeroing on read failures
- Success/failure statistics tracking per sensor
- Synchronized timestamp assignment for data correlation
- FreeRTOS task implementation for real-time operation

The implementation ensures data freshness through timeout checks and maintains consistent sampling across all sensors within the same timeframe.

(TODO):
Testing:
- Verify initialization of all IMU drivers
- Confirm 5ms sampling period is maintained
- Check error handling when sensor reads fail
- Validate data flow to estimator module


src/application/imu_handler/imu_handler.c
```diff
@@ -1 +1,227 @@
...
+        uint32_t success_count;
+        uint32_t failure_count;
+    } polulu_stats, st_stats, movella_stats;
+} imu_handler_state_t;
+
+static imu_handler_state_t imu_handler_state = {0};
+
+/**
+ * @brief Initialize all IMU hardware
+ * @return Status of the initialization operation (success only if all IMUs initialize)
```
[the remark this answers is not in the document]
Joe-Joe-Joe-Joe: I see you commented on the other init to run it before scheduler. still good to be verbose here as well

src/application/imu_handler/imu_handler.h
```diff
@@ -1,8 +1,26 @@
...
+ * @return Status of initialization
+ */
+w_status_t imu_handler_init(void);
+
+/**
+ * @brief Execute one iteration of the IMU handler processing
+ * Reads data from all IMUs and updates the estimator
+ * @return Status of the execution
+ */
+w_status_t imu_handler_run(void);
```
Joe-Joe-Joe-Joe, line 18: I was going back and forth w @celery6 on this. I don't think we should expose functions like this in headers, because they can cause unintended behaviour if run outside the intended task.

Either wrap in `#ifdef GTEST` so it is only included in the header for unit testing, or pass static members as parameters instead of accessing them from the outer scope. So like `w_status_t imu_handler_run(imu_handler_state_t *state);`. I tend to prefer the latter but in this case it also affects state est so ifdefs might be more reliable.

tests/mocks/mock_altimu_10.h
Joe-Joe-Joe-Joe, line 1: you don't need to write mock headers for modules, you can just include the real header in the mock source and the unit test file

tests/mocks/mock_freertos.c
```diff
@@ -1,7 +1,16 @@
 #include "mock_freertos.h"
+#include <stdlib.h>
 
-// Define mock here
-DEFINE_FAKE_VOID_FUNC(vTaskDelay, uint32_t);
+// After one iteration, exit the task
+static void vTaskDelay_custom_fake(TickType_t xTicksToDelay) {
+    (void)xTicksToDelay;
+    exit(0); // Cleanly exit after first iteration
```
Joe-Joe-Joe-Joe, line 7: sus? we definitely should not be calling `exit` ever, right? Is this like a unit testing-specific thing?

Joe-Joe-Joe-Joe (changes requested):

tests/mocks/mock_freertos.c
taisirhassan: ya this is for unit testing. we're using a custom fake for one iteration cuz we don't want our task to be stuck in an infinite loop when unit testing lol

src/application/imu_handler/imu_handler.c
```diff
@@ -1 +1,228 @@
...
+    // Set timestamps for all IMUs
+    // Note: All IMUs get the same timestamp intentionally for synchronization
+    imu_data.polulu.timestamp_imu = now_ms;
+    imu_data.st.timestamp_imu = now_ms;
+    imu_data.movella.timestamp_imu = now_ms;
+
+    // Read from all IMUs
+    read_pololu_imu(&imu_data.polulu);
+    read_st_imu(&imu_data.st);
+    read_movella_imu(&imu_data.movella);
```
celery6, line 194: its required to read the w_stauts_t return vals from these and check for err

src/application/imu_handler/imu_handler.c
```diff
@@ -1 +1,228 @@
...
+/**
+ * @brief IMU handler task function for FreeRTOS
+ * @note This task will be created during system initialization
+ * @param argument Task argument (unused)
+ */
+void imu_handler_task(void *argument) {
+    (void)argument; // Unused parameter
+
+    // Initialize all IMUs - must be done after scheduler start
+    initialize_all_imus();
```
celery6, line 219: check status. should probably add todo to handle this error fatally or smth cuz u cannot run the rest of the task if this fails..

src/application/imu_handler/imu_handler.c
```diff
@@ -1 +1,228 @@
...
+
+    // Main task loop
+    while (1) {
+        imu_handler_run();
```
celery6, line 223: check status probably?

src/application/imu_handler/imu_handler.c
```diff
@@ -1 +1,228 @@
...
+
+        // Wait for next sampling period
+        vTaskDelay(pdMS_TO_TICKS(IMU_SAMPLING_PERIOD_MS));
```
celery6, line 226: use xtaskdelayuntil, and check the return value to ensure delay didnt get overrun. google it

src/application/imu_handler/imu_handler.c
celery6, line 1: add is_dead, and delete all references to ST imu 😔

celery6 (changes requested):

CMakeLists.txt
```diff
@@ -11,6 +11,11 @@ set(CMAKE_PROJECT_NAME processor_board)
 set(BUILD_TARGET "firmware" CACHE STRING "Select the build target: firmware or test")
 set_property(CACHE BUILD_TARGET PROPERTY STRINGS firmware test)
 
+# Enable parallel builds using all available cores unconditionally
+include(ProcessorCount)
+ProcessorCount(N)
+set(CMAKE_BUILD_PARALLEL_LEVEL ${N})
+
```
celery6, line 18: pretty sure ninja uses all cores by default .. but ok

src/application/imu_handler/imu_handler.c
```diff
@@ -1 +1,223 @@
...
+    // Read barometer data
+    altimu_barometer_data_t baro_data;
+    status |= altimu_get_baro_data(&baro_data);
+
+    if (W_SUCCESS == status) {
+        imu_data->barometer = baro_data.pressure;
+        imu_data->is_dead = false;
+        imu_handler_state.polulu_stats.success_count++;
+    } else {
+        // Zero all data if any reading fails to meet requirements
```
celery6, line 81: keep in mind for later optimization maybe if needed: as long as is_dead is set, it doesnt matter if the values are 0 or anything. so memset is technically unecessary

src/application/imu_handler/imu_handler.h
```diff
@@ -1,9 +1,30 @@
...
 
+#ifdef GTEST
+/**
+ * @brief Execute one iteration of the IMU handler processing
+ * Reads data from all IMUs and updates the estimator
+ * @note This function is exposed only for unit testing
+ * @return Status of the execution
+ */
+w_status_t imu_handler_run(void);
+#endif
```
celery6, line 28: i think u can `extern` the function declaration directly inside the unit test file since it's not a static func? next time

tests/CMakeLists.txt
```diff
@@ -1,9 +1,35 @@
...
+
+# Add coverage flags - safely to avoid libgcov errors
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0 -Wno-unused-parameter")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -O0 -Wno-unused-parameter")
+
+# Add linker flags for gcov
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -Wl,--no-as-needed -Wl,--gc-sections")
+
+# Set coverage flags per-file to avoid conflicts - add atomic update to prevent timestamp errors
+set(COVERAGE_COMPILE_FLAGS "--coverage -fprofile-arcs -ftest-coverage -fprofile-correction -fprofile-update=atomic")
```
celery6, line 15: ORZZZZZZZZZZZ ORZOROZROZROORZOORORZ DEVOPS PROFESIONELASFASDJALKGSDLKF

celery6 (approved):

tests/unit/imu_handler_test.cpp
```diff
@@ -0,0 +1,355 @@
...
+#include "application/imu_handler/imu_handler.h"
+#include "common/math/math.h"
+#include "drivers/altimu-10/altimu-10.h"
+#include "drivers/movella/movella.h"
+#include "drivers/timer/timer.h"
+#include "third_party/rocketlib/include/common.h"
+#include "mock_freertos.h"
+
+    // Forward declare imu_handler_run
+extern w_status_t imu_handler_run(void);
```
Joe-Joe-Joe-Joe, line 16: Functions are linked externally by default, so I don't think this is necessary

Joe-Joe-Joe-Joe (approved):
LGTM

hope it works IRL!

tests/unit/imu_handler_test.cpp
celery6: u will get an undeclared function warning I think? 
(though it would still run yes)

PR #146, opened 2025-02-27
Math: algebra for vector3d_t

FinnBreu:
(Formerly math wrapper, now just custom type math)

We use the vector3d_t type a lot.
IMU handler, Estimator models, (Controller) all do math operations with this type.
These should be all the needed math functions


Redstone-ray (approved):

src/common/math/math-wrapper.h
```diff
@@ -0,0 +1,62 @@
...
+vector3d_t math_vector_rotate(matrix3d_t matrix, vector3d_t vector_right);
+
+// matrix ^T | transpose, only in R3x3 for inverting rotation matrices
+matrix3d_t math_matrix_transp3(matrix3d_t input);
+
+/**
+ * General operations
+ */
+// vector * scalar | vector scaling
+void math_vector_scale(float *result, float *vector, float scalar);
```
Joe-Joe-Joe-Joe, line 26: Am I dumb, don't you need to specify the length of your vectors for basically all these general operations?

FinnBreu: Just `(sizeof vector) / (sizeof *vector)` inside the function should work?

FinnBreu: Ughh I miss matlab

celery6: would it be more or less convenient to make a vector struct and matrix struct type that holds the length so u dont have to pass it into every function

FinnBreu: thats just using the DSP syntax directly

src/common/math/math-algebra3d.c
```diff
@@ -0,0 +1,88 @@
+#include "common/math/math-algebra3d.h"
+#include "common/math/math.h"
+#include <stdbool.h>
+#include <stdint.h>
+#include <math.h>
```
Joe-Joe-Joe-Joe, line 5: is this necessary?

src/common/math/math-algebra3d.c
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Special operations in R3, as they are used often
+ * Cross product only exists in R3 
+ * Rotation matrix transpose is inverse rotation
+ */
+// vector * scalar | vector scaling
```
Joe-Joe-Joe-Joe, line 12: I would avoid `|` since it means something specific in C. Would use `->` but that technically _also_ means something specific in C. Could be very explicit and use doxygen style, idk

src/common/math/math-algebra3d.c
```diff
@@ -0,0 +1,88 @@
...
+vector3d_t math_vector_scale(float scalar, const vector3d_t *vector)
```
Joe-Joe-Joe-Joe, line 13: It's very obvious within the source file, but when the functions are actually being used its not immediately clear they are 3d vector operations. Would prefer `math_vector3d_scale` or similar

src/common/math/math-algebra3d.c
```diff
@@ -0,0 +1,88 @@
...
+vector3d_t math_vector_cross(const vector3d_t *vector_left, const vector3d_t *vector_right)
+{
+    vector3d_t result;
+    result.x = vector_left->y * vector_right->z - vector_left->z * vector_right->y;
+    result.y = vector_left->z * vector_right->x - vector_left->x * vector_right->z;
+    result.z = vector_left->x * vector_right->y - vector_left->y * vector_right->x;
+    return result;
+}
+
+/* // matrix * vector | matrix vector multiplication, for rotating vectors
```
Joe-Joe-Joe-Joe, line 60: remove?

Joe-Joe-Joe-Joe (changes requested):
You are failing the formatting check. Not sure how as VScode should auto-format on save

Run `scripts/format.sh` from the root directory in the container if necessary

src/common/math/math-algebra3d.c
Joe-Joe-Joe-Joe: This is fine if we are only operating on 3d vectors in the entire code but I assume we aren't (e.g. estimator state vector)

celery6:
> You are failing the formatting check. Not sure how as VScode should auto-format on save
> 
> Run `scripts/format.sh` from the root directory in the container if necessary

I think formatter ignores .c files in the src/common folder cuz there used to be none. I will fix that. or u can fix that by editing the clang format file

celery6:
actually by editing the format.sh file *

src/common/math/math-algebra3d.c
```diff
@@ -0,0 +1,75 @@
+/**
+ * Special operations in R3, as they are used often
+ * Cross product only exists in R3
+ * Rotation matrix transpose is inverse rotation
+ */
+#include "common/math/math-algebra3d.h"
+#include "common/math/math.h"
+
+// vector * scalar // vector scaling
+vector3d_t math_vector3d_scale(float scalar, const vector3d_t *vector)
```
Joe-Joe-Joe-Joe, line 10: technically none of these need to be pointers

Joe-Joe-Joe-Joe (approved):
orz

src/common/math/math-algebra3d.c
FinnBreu: That was the intention.
But also on that thread:

> "If you're passing or returning structs by value, copies of those structs will get placed on the stack" I'd call braindead any toolchain that does so. Yes, it's sad that so many will do it, but it's not anything that the C standard calls for. A sane compiler will optimize it all out

@celery6 what does our toolchain do

Redstone-ray (approved):

src/common/math/math-algebra3d.c
celery6: idk?? I think the comment after that addresses it though.. @KubaOber This is why that doesn't get done often: stackoverflow.com/questions/552134/… 

https://stackoverflow.com/questions/59976295/structs-and-unions-which-is-better-from-a-performance-point-of-view-passing-th
and this shows its almost certainly better to pass by pointer instead of value.

so to be safe let's just use pointer instead of value, and if anyone wants to do a deep dive then that's a later to-do

celery6 (approved):

Joe-Joe-Joe-Joe (approved):

PR #149, opened 2025-03-03
Controller gain interpolation

SAT-oO:


src/application/controller/controller.c
```diff
@@ -11,16 +16,30 @@ TickType_t timeout = pdMS_TO_TICKS(5);
...
 typedef struct {
     float gain_arr[GAIN_NUM];
-    FILE *gain_table;
-    vector3d_t gain_k;
+    arm_bilinear_interp_instance_f32 gain_instance;
+
+    float gain_k[3];
     float gain_k_pre;
     float reference_signal;
 } gain_table_entry_t;
```
FinnBreu, line 24: What is this struct????

Why is there gain_array[4] AND gain_k[3] with gain_k_pre? What not use a union, similar to the quaternion_t from common math?

Why is an interpolation instance here, should that not hold be a reference to the entire table instead of one entry?

Why is there a reference signal float here? that has nothing to do with the gains!

src/application/controller/controller.c
```diff
@@ -11,16 +16,30 @@ TickType_t timeout = pdMS_TO_TICKS(5);
...
+// placeholder: suppose this is imported daa
+// for normalizing inputs
+float canard_coeff_scale;
+float canard_coeff_offset;
+float pressure_dynamic_scale;
+float pressure_dynamic_offset;
+// for building each interpolation instance
+const uint16_t numRow;
+const uint16_t numCol;
+static float gain_table[GAIN_NUM][10000];
```
FinnBreu, line 35: Should be very doable to output this (but with values) from matlab

src/application/controller/controller.c
```diff
@@ -129,38 +138,80 @@ void controller_task(void *argument) {
...
-                }else if(timestamp > 26){
-                    gain_table_entry.reference_signal = 0;
+            uint32_t time = controller_state.current_state.timestamp - TIME_START;
+            if (time > 5) {
+                if (time < 12) {
+                    gain_entry.reference_signal = 1;
+                } else if (time < 19) {
+                    gain_entry.reference_signal = -1;
+                } else if (time > 26) {
+                    gain_entry.reference_signal = 0;
```
FinnBreu, line 152: Again, gain_entry.reference makes no sense. 
reference_signal should just be a float, not in a struct

src/application/controller/controller.c
```diff
@@ -129,38 +138,80 @@ void controller_task(void *argument) {
...
+                controller_state.current_state.roll_rate,
+                controller_state.current_state.canard_angle
+
+            };
+
+            arm_dot_prod_f32(
+                gain_entry.gain_k, roll_state, FEEDBACK_GAIN_NUM, &controller_output.commanded_angle
+            );
+            controller_output.commanded_angle +=
+                gain_entry.reference_signal * gain_entry.gain_k_pre;
```
FinnBreu, line 199: Looks good

src/application/controller/controller.c
```diff
@@ -129,38 +138,80 @@ void controller_task(void *argument) {
...
+                }
+            }
+
+            // TODO compute control output and overwrites output queue
+            float roll_state[3] = {
+                controller_state.current_state.roll_angle,
+                controller_state.current_state.roll_rate,
+                controller_state.current_state.canard_angle
+
+            };
```
FinnBreu, line 193: Fine, but seems more convoluted than necessary. Would it be easier to define the roll state as a union, that way it doesn't need to be copied here?

FinnBreu (changes requested):

src/application/controller/controller.c
```diff
@@ -129,38 +138,80 @@ void controller_task(void *argument) {
...
-            if(timestamp > 5){
-                if(timestamp < 12){
-                    gain_table_entry.reference_signal = 1;
-                }else if(timestamp < 19){
-                    gain_table_entry.reference_signal = -1;
-                }else if(timestamp > 26){
-                    gain_table_entry.reference_signal = 0;
+            uint32_t time = controller_state.current_state.timestamp - TIME_START;
+            if (time > 5) {
+                if (time < 12) {
```
FinnBreu, line 147: We should probably make this roll program here stand out more. Maybe put it at the top of the function, or make the time stamp threshold some constants defined on top, or both.
Right now these are magic numbers hidden in the middle of the file

FinnBreu (changes requested):

src/application/controller/controller.c
SAT-oO: shoved everything here just for organization, will change that 😶‍🌫️

src/application/controller/controller.c
SAT-oO: yep code will look cleaner that way, will change that too

src/application/controller/controller.c
SAT-oO: 👌

src/application/controller/controller.c
```diff
@@ -11,16 +16,30 @@ TickType_t timeout = pdMS_TO_TICKS(5);
...
-    FILE *gain_table;
-    vector3d_t gain_k;
+    arm_bilinear_interp_instance_f32 gain_instance;
+
+    float gain_k[3];
     float gain_k_pre;
     float reference_signal;
 } gain_table_entry_t;
 
+// placeholder: suppose this is imported daa
```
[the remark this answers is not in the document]
Joe-Joe-Joe-Joe: That we will then overwrite with the converted matlab

Joe-Joe-Joe-Joe (changes requested):

Joe-Joe-Joe-Joe:
See #116

Issue #80, opened 2025-01-29
design, implement initialization routine


celery6:
last year we simply called every module init function + taskcreate in main.c. That had NONEXISTANT err handling. We need to change it so the init process can try again repeadetly if it fails. Also look into maybe move taskcreate calls into each module's init? I don't see why it should be separate since a modules task is the modules responsibility

PR #153, opened 2025-03-05
Sd card header file & implementation

Wehbucci:
This PR introduces the SD card module as a wrapper around FATFS, providing thread-safe file operations for the  firmware. The module includes essential functionalities such as file creation, reading, writing, deletion, and status checking, ensuring safe access through a mutex-protected interface.


Joe-Joe-Joe-Joe:
Does this build for you locally?

Joe-Joe-Joe-Joe:
first thing you should do is merge main into your branch

src/drivers/sd_card/sd_card.c
```diff
@@ -0,0 +1,183 @@
+#include "sd_card.h"
```
Joe-Joe-Joe-Joe, line 1:
```suggestion
#include "drivers/sd_card/sd_card.h"
```
source include paths should use path relative to `src/`

src/drivers/sd_card/sd_card.h
```diff
@@ -1,6 +1,9 @@
 #ifndef SD_CARD_H
 #define SD_CARD_H
 
+#include "fatfs.h"
+#include "semphr.h"
```
Joe-Joe-Joe-Joe, line 5: These two headers aren't necessary for header declarations, so they should be moved into the .c file

Joe-Joe-Joe-Joe (changes requested):
I think the reason this compiles locally is due to an outdated build config. I think if you fix the paths and merge the latest version of main, your pipeline should pass. I don't see where the stm32 autogenerated headers are included, so I don't know if you need to change the include path for `fatfs.h`. Other files seem to include them directly, so keep them for now and see if it builds.

Wehbucci:
it fails to build locally now, ignore the failed pipeline, I'll try and fix it

src/drivers/sd_card/sd_card.c
```diff
@@ -1 +1,186 @@
 #include "drivers/sd_card/sd_card.h"
+#include "../Core/Inc/sdmmc.h"
```
celery6, line 2: is this the right include or should it be include stm32h7xx_hal like everything else uses..? there's a note abt that in the practical coding advice doc but idk if sd card is different

src/drivers/sd_card/sd_card.c
```diff
@@ -1 +1,186 @@
...
+        return W_FAILURE;
+    }
+
+    return W_SUCCESS;
+}
+
+w_status_t file_read(char *fileName, void *buffer, uint32_t bufferSize, uint32_t *bytesRead)
+{
+    /* Ensure thread-safe access to the SD card. */
+    if (xSemaphoreTake(sd_mutex, portMAX_DELAY) != pdTRUE)
```
celery6, line 38: don't use portmaxdelay because u don't want this to freeze the entire program forever if sd card died.
either add a param to the function to let users pass in a max timeout, or set this to 0

celery6: commemt applies to all the portmaxdelays in here

celery6 (changes requested):

celery6:
f_mount doesnt work. everything else works. will comment out file_delete because we definitely do not need it and dont want to accidentally call it

celery6:
`fmount(..., ..., 1)` doesnt work if its called after any other freertos init occurs. because calling any freertos api before scheduler start masks all interrupts, including systick? https://github.com/FreeRTOS/FreeRTOS-Kernel/issues/254#issuecomment-769904807
and fmount depends on hal_delay, which depends on timer interrupt working. so if all interrupts r masked, fmount hangs forever.

this is irrelevant if use `fmount(.., .., 0)` to avoid actually mounting it (hence avoid using hal delay).

so .. we should wait until after scheduler start to mount this? i think this came up at some point last year but no one questioned it 🤔  but also we used `fmoun(..., ..., 0)` so it prob didnt matter anyway

Core/Src/freertos.c
```diff
@@ -51,9 +51,9 @@
 /* Definitions for defaultTask */
 osThreadId_t defaultTaskHandle;
 const osThreadAttr_t defaultTask_attributes = {
-    .name = "defaultTask",
-    .stack_size = 512 * 4,
-    .priority = (osPriority_t)osPriorityNormal,
+  .name = "defaultTask",
+  .stack_size = 128 * 4,
```
Joe-Joe-Joe-Joe, line 55: Will this stackoverflow if I printf from default task? (I do that a lot)

Core/Src/main.c
```diff
@@ -146,7 +149,7 @@ int main(void) {
 
     BaseType_t status2 = pdTRUE;
     status2 &=
-        xTaskCreate(flight_phase_task, "flightphase", 512, NULL, 1, &flight_phase_task_handle);
+        xTaskCreate(flight_phase_task, "flightphase", 2048, NULL, 1, &flight_phase_task_handle);
```
Joe-Joe-Joe-Joe, line 152: What is happening to our stack sizes?

FATFS/Target/ffconf.h
```diff
@@ -261,10 +258,11 @@
 /  SemaphoreHandle_t and etc.. A header file for O/S definitions needs to be
 /  included somewhere in the scope of ff.h. */
 
-/* define the ff_malloc ff_free macros as FreeRTOS pvPortMalloc and vPortFree macros */
+/* define the ff_malloc ff_free macros as standard malloc free */
```
Joe-Joe-Joe-Joe, line 261: malloc sus???

cansw_processor_canards.ioc
```diff
@@ -27,16 +29,18 @@ FDCAN1.RxFifo0ElmtsNbr=64
...
-FREERTOS.INCLUDE_xEventGroupSetBitFromISR=1
-FREERTOS.IPParameters=Tasks01,configENABLE_FPU,INCLUDE_xEventGroupSetBitFromISR,configUSE_NEWLIB_REENTRANT,FootprintOK,configTOTAL_HEAP_SIZE,configCHECK_FOR_STACK_OVERFLOW,configTIMER_TASK_PRIORITY,configENABLE_BACKWARD_COMPATIBILITY
-FREERTOS.Tasks01=defaultTask,24,512,StartDefaultTask,Default,NULL,Dynamic,NULL,NULL
-FREERTOS.configCHECK_FOR_STACK_OVERFLOW=1
-FREERTOS.configENABLE_BACKWARD_COMPATIBILITY=0
+FREERTOS.IPParameters=Tasks01,configENABLE_FPU,configRECORD_STACK_HIGH_ADDRESS,configCHECK_FOR_STACK_OVERFLOW,configGENERATE_RUN_TIME_STATS,configUSE_OS2_MUTEX,configUSE_OS2_TIMER,configUSE_OS2_THREAD_FLAGS,configUSE_OS2_EVENTFLAGS_FROM_ISR,configUSE_OS2_THREAD_ENUMERATE,configUSE_OS2_THREAD_SUSPEND_RESUME
+FREERTOS.Tasks01=defaultTask,24,128,StartDefaultTask,Default,NULL,Dynamic,NULL,NULL
+FREERTOS.configCHECK_FOR_STACK_OVERFLOW=2
 FREERTOS.configENABLE_FPU=1
-FREERTOS.configTIMER_TASK_PRIORITY=50
```
Joe-Joe-Joe-Joe, line 37: This is extremely important, please do not change it

src/drivers/sd_card/sd_card.h
```diff
@@ -9,21 +9,24 @@
...
 
 /**
  * @brief Read a file from the SD card.
  *
- * @param fileName - The name of the file to read.
+ * @param file_name - The name of the file to read.
  * @param buffer - The buffer to read the file into.
- * @param bufferSize - The size of the buffer.
- * @param bytesRead - The number of bytes read from the file.
+ * @param num_bytes - Number of bytes t read from the file.
```
Joe-Joe-Joe-Joe, line 24: *to

src/drivers/sd_card/sd_card.h
```diff
@@ -9,21 +9,24 @@
...
+ * @param bytes_read - The number of bytes read from the file.
```
Joe-Joe-Joe-Joe, line 25: I think "bytes successfully read" might be more clear

src/drivers/sd_card/sd_card.c
```diff
@@ -1 +1,182 @@
 #include "drivers/sd_card/sd_card.h"
+#include "FreeRTOS.h"
+#include "fatfs.h"
+#include "semphr.h"
+#include "stm32h7xx_hal.h"
+
+extern SD_HandleTypeDef hsd1;
```
Joe-Joe-Joe-Joe, line 7: Use project convention of passing to init function rather than externing

src/drivers/sd_card/sd_card.c
```diff
@@ -1 +1,182 @@
...
+    if (f_mount(&g_fs_obj, "", 1) != FR_OK) {
+        return W_FAILURE;
+    }
+
+    /*
+     * Create a mutex for thread-safe SD card operations.
+     * This mutex will ensure that concurrent tasks do not access the SD card simultaneously,
+     * which helps prevent data corruption.
+     */
+    sd_mutex = xSemaphoreCreateMutex();
```
Joe-Joe-Joe-Joe, line 35: If a task attempts to take the mutex before it is created, the code will die. If we can't call this init until after scheduler start, then there needs to be some sort of exclusion mechanism. This should be a requirement for any module that creates freertos objects after scheduler start.
image

Joe-Joe-Joe-Joe (changes requested):
Does `fmount` now work irl also?

cansw_processor_canards.ioc
celery6: yeah i spent like 3 hr reading about freertos timer priority and whatnot
anyway why tf did this change in the .ioc I HATE CUBEMX
image

FATFS/Target/ffconf.h
celery6: apparently this is only relevant if u enable Long File Name (LFN)?
image
image

ctrl+f the whole project shows no actual uses of fatfs malloc if it did theres bigger problems 💀

cansw_processor_canards.ioc
celery6: ok its back at 55 (aka maxpriority - 1) in the .ioc but doesnt show here maybe thats the default idk

src/drivers/sd_card/sd_card.c
celery6: all this init behaviour is being figured out in #80 
idt thats the responsibility of sd card module? if u dont init before using then dying is indeed expected?

src/drivers/sd_card/sd_card.h
celery6: how can u even "unsuccessfully" read a byte though

src/drivers/sd_card/sd_card.c
celery6: done

src/drivers/sd_card/sd_card.c
Joe-Joe-Joe-Joe: yes, but who is initing? Is every task that calls an sd card function expected to call init first? If so, that's not clear. Does multiple inits cause issues? Seems like it would. Its not protected by the mutex, so one task can remount while another is doing a R/W
image


I stand by my earlier statement. If init cannot be called in the big block before scheduler start, it should have a flag that prevents other SD card functions from being executed prior to init, and should prevent multiple calls to init

src/drivers/sd_card/sd_card.h
Joe-Joe-Joe-Joe: you ask for 20 bytes but there are only 10?

its a nit. `bytes to read from the file` and `bytes read from the file` are very similar sentences because english moment, which makes them harder to parse. I think adding an extra word in there makes the second sentence more distinct and the intent more clear. up to you

Joe-Joe-Joe-Joe (changes requested):
Please do not mass edit the PLL configs
This will, among other things, make our execution twice as slow
image

src/drivers/sd_card/sd_card.c
celery6: https://waterloorocketry.slack.com/archives/C05EUQLFP4J/p1742430255861739
resolved?

martendo (approved):

martendo:
I still might want the ability to overwrite a file but due to the issues I've been experiencing I'm still debating about whether or not this is the best approach, for now LGTM

PR #154, opened 2025-03-07
Update processor_board_Debug.launch

celery6:
oops updated cmake name


Joe-Joe-Joe-Joe (approved):
LGTM i can't be bothered to flash a board to test this its probably fine

celery6:
currently sitting in the bay so yeah it works

PR #157, opened 2025-03-07
Movella implementation

### LINKED ISSUE #82: implement movella driver

Joe-Joe-Joe-Joe:
closes #82


src/drivers/uart/uart.c
```diff
@@ -48,9 +48,11 @@ static uart_stats_t s_uart_stats[UART_CHANNEL_COUNT] = {0};
  * @brief Initialize UART channel
  * @param channel UART channel to initialize
  * @param huart HAL UART handle
+ * @param timeout_ms Operation timeout in milliseconds
```
celery6, line 51: we decided to delete the uart module `timeout` param cuz each read/write passes in a timeout instead. remove all of these? not urgent though since timeout variable is doing nothing anyway

celery6 (approved):

src/drivers/uart/uart.c
Joe-Joe-Joe-Joe: Would change but I'm paranoid about inadvertently breaking the code again

PR #158, opened 2025-03-09
assorted devops improvements and add imu gpio pins to gpio module

celery6:


Joe-Joe-Joe-Joe (approved):
LGTM
i don't know what most of this devops stuff does but I trust you!

PR #163, opened 2025-03-13
Extract flight phase update function

Joe-Joe-Joe-Joe:
As requested


Joe-Joe-Joe-Joe:
Coverage is still kinda bad since we can't hit the callbacks. I should note that the default case is difficult to test since the compiler is very good at noticing when I try to pass something that isn't a valid state to the update function

celery6:
use casts to bypass compiler warnings (pls try to not do that in real code though:face)

Joe-Joe-Joe-Joe:
Yeah I tried casting a pointer to an into to a pointer to a flight_phase_state_t and the compiler catches that too

src/application/flight_phase/flight_phase.h
```diff
@@ -57,4 +57,8 @@ w_status_t flight_phase_send_event(flight_phase_event_t event);
  */
 w_status_t flight_phase_reset(void);
```
celery6, line 59: u don't need to declare this here even for unit testing. just use extern in the test file to access it. idk if that's better practice but it avoids polluting this source file with ifdefs

celery6:
when mergeeee

Joe-Joe-Joe-Joe:
I'm descoping this, implement as you see fit

taisirhassan (approved):

PR #170, opened 2025-03-20
all ocnfigured ioc hopefully

celery6:
will test all peripherals are reachable in bare metal code. then never touch this again fml

- remove timer1 cuz unused
- enable cordic and fmac interrupts for future use if interested
- freertos timer task priority 55 (max - 1) as recommended by many sources
- disable fatfs reentrancy because we do not access sd card from 2 threads at once (mutex protected)


celery6:
why does flight phase need to be highest

Joe-Joe-Joe-Joe (approved):
:shipit:

PR #171, opened 2025-03-21
OLD 2025 - HIL Test Harness

punchdii:


Core/Src/main.c
```diff
@@ -116,6 +117,10 @@ int main(void) {
     /* USER CODE BEGIN 2 */
     HAL_TIM_Base_Start(&htim2);
 
+    // !!! HIL TESTING MODIFICATION: Disable SysTick before starting scheduler !!!
+    // The RTOS tick will be driven by hil_increment_tick() called from UART RX ISR.
+    SysTick->CTRL &= ~SysTick_CTRL_ENABLE_Msk;
```
celery6, line 122: are u sure this mask persists through the freertos kernel init? pretty sure freertos unmasks interrupts immediately before entering the first task ... ?????

celery6: why not just edit the systick isr to be blank (comment out the tickincrement call in there)

src/CMakeLists.txt
```diff
@@ -82,11 +85,11 @@ add_executable(proc_binary)
 
 # get processor source files
 file(GLOB_RECURSE PROC_SOURCES
-    "application/*.c"
+    "application/[!h][!i][!l]*/*.c" # Exclude hil from this glob
```
celery6, line 88: i lol'd at this glob this looks so funny :FAcCECee

src/CMakeLists.txt
```diff
@@ -82,11 +85,11 @@ add_executable(proc_binary)
...
     "drivers/*.c"
 )
 
-# add thr source files
+# add the source files
```
celery6, line 92: kys

src/application/hil/simulator.c
```diff
@@ -0,0 +1,149 @@
...
+#include "FreeRTOS.h"
+#include "application/estimator/estimator.h"
+#include "task.h"
+#include <string.h>
+
+// Define the expected payload size based on the document
+#define SIM_EXPECTED_PAYLOAD_SIZE 140
+
+// Define payload offsets based on the document
+enum simulator_data_offsets {
```
celery6, line 16: ... if ur gonna define a type for the incoming packet can u please create a `packed` struct like
struct sim_data_packet {
 float encoder_Angle,
float accelx,
...
}
and cast the incoming data to this type. why use offset funny pointer arithmetic when it's a well-defined structure 😭

src/application/hil/CMakeLists.txt
celery6, line 1: why is this separate cmakelist needed ... ??? overcomplicating much
please minimize complexity of the hil harness as much as possible

src/application/hil/simulator.c
```diff
@@ -0,0 +1,149 @@
...
+    *canard_angle = latest_canard_angle_rad;
+    new_control_output_available = false; // Mark as consumed
+    return W_SUCCESS;
+}
+
+/**
+ * @brief Set the latest control output (called by the controller task).
+ */
+void simulator_set_control_output(float canard_angle) {
+    // Atomically update the latest angle (not strictly necessary with one writer)
```
celery6, line 144: huh who else is sending canard cmds ?

src/application/hil/simulator.h
celery6, line 1: why is this in a separate file and not all in hil.c

src/drivers/uart/uart.c
```diff
@@ -5,14 +5,26 @@
...
 #include <stdint.h>
+
 #include <string.h>
+
 /* Static buffer pool for all channels */
 static uint8_t s_buffer_pool[UART_CHANNEL_COUNT][UART_MAX_LEN * UART_NUM_RX_BUFFERS];
 
+// HIL Constants (needed for framing the response)
+#define HIL_UART_HEADER_CHAR '?'
+#define HIL_UART_FOOTER_CHAR '\n'
```
celery6, line 26: these are defined in hil.c too

src/drivers/uart/uart.c
```diff
@@ -216,53 +224,58 @@ void HAL_UARTEx_RxEventCallback(UART_HandleTypeDef *huart, uint16_t size) {
...
+        // Normal non-HIL operation
+        estimator_all_imus_input_t imu_data = {0};
+        if (handle->package_counter % 5 == 0) {
+            // ... (Existing non-HIL data processing code) ...
+        }
     }
+    */
+
     // Advance to next buffer in circular arrangement
     uint8_t next_buffer = (curr_buffer + 1) % UART_NUM_RX_BUFFERS;
```
celery6: we discussed that for HIL, there does not need to be buffer rotation because u only read+process 1 packet at a time? is this right? removing buffer rotation just helps reduce complexity for testing

celery6 (changes requested):

Core/Src/main.c
taisirhassan: to answer your two points: The SysTick_Handler does two things:
one, it calls HAL_IncTick() which increments a counter used by the STM32 HAL library for its internal timing (for ex like timeouts in HAL_UART_Transmit, HAL_Delay, etc.).
two, it calls osSystickHandler() which leads to xTaskIncrementTick() and thhis increments the FreeRTOS tick count for task scheduling.

if we were to completely blank out SysTick_Handler or just comment out HAL_IncTick(), many standard HAL functions that rely on timeouts will most definitely break. we still want the hardware peripherals and their basic HAL drivers to function based on real hardware time, even though we're manipulating the RTOS's sense of time for scheduling.

for your first point: yeah it wouldn't persist throughout the entire kernel init only until hil_increment_tick returns prolly will just define a macro and wrap it around the systick handler to fix this

src/drivers/uart/uart.c
taisirhassan: meh replace it with a linear buffer? just sounds like a pain i'd rather keep it shouldn't be too complex

celery6:
bye old hil hello new hil #567  🥵

PR #172, opened 2025-03-23
Controller calc -- gain interpolation/canard angle compute 

SAT-oO:


SAT-oO:
see the gain interp commit for actual code instead of build fix 💀  (https://github.com/waterloo-rocketry/cansw_processor_canards/pull/172/commits/82a74939e00380c3a5a2bf364b51058d8610f2f8)

src/application/controller/gain_scheduling.h
```diff
@@ -17,18 +20,10 @@ typedef union {
...
-extern float canard_coeff_scale;
-extern float canard_coeff_offset;
-extern float pressure_dynamic_scale;
-extern float pressure_dynamic_offset;
-// for building each interpolation instance
-extern const uint16_t numRow;
-extern const uint16_t numCol;
-static float gain_table[GAIN_NUM][6000] __attribute__((unused));
+// coordinate conversion
+extern float interpolate_gain(float *p_dyn, float *coeff, int gain_index);
```
Joe-Joe-Joe-Joe, line 27: Functions are `extern` by default

src/application/controller/gain_scheduling.h
```diff
@@ -17,18 +20,10 @@ typedef union {
 
 } controller_gain_t;
 
-extern arm_bilinear_interp_instance_f32 gain_instance;
 static const float max_commanded_angle = 20 * 180.0 / M_PI;
+static const float reference_signal = 0.0f; // no roll program for test flight
```
Joe-Joe-Joe-Joe, line 24: Why not declare in `controller.c`?

Joe-Joe-Joe-Joe (changes requested):
- Can you just delete all these extraneous files?
- I can't add inline suggestions to code you didn't change, but CAN handler is now merged so you certainly have everything you need to do it. I think I mentioned in prior PR but I can explain what specific functions you need for this
~~- Obviously, my comments on the matlab-generated part of gain table need to be addressed in matlab first~~
- I now realize the interpolation function was written after. I think the matlab generated portion is fine.

src/application/controller/gain_scheduling.c
```diff
@@ -1 +1,3472 @@
...
+     -4.1475E-01, -4.2153E-01, -4.3009E-01, -4.4139E-01, -4.5740E-01, -4.8271E-01, -5.3231E-01,
+     -7.1518E-01, 7.1518E-01,  5.3231E-01,  4.8271E-01,  4.5740E-01,  4.4139E-01,  4.3009E-01,
+     4.2153E-01,  4.1475E-01,  4.0919E-01,  4.0451E-01,  4.0049E-01,  3.9699E-01,  3.9390E-01,
+     3.9113E-01,  3.8864E-01,  -3.8130E-01, -3.8373E-01, -3.8643E-01, -3.8945E-01, -3.9286E-01,
+     -3.9678E-01, -4.0134E-01, -4.0676E-01, -4.1337E-01, -4.2170E-01, -4.3271E-01, -4.4828E-01,
+     -4.7290E-01, -5.2112E-01, -6.9870E-01, 6.9870E-01,  5.2112E-01,  4.7290E-01,  4.4828E-01,
+     4.3271E-01,  4.2170E-01,  4.1337E-01,  4.0676E-01,  4.0134E-01,  3.9678E-01,  3.9286E-01,
+     3.8945E-01,  3.8643E-01,  3.8373E-01,  3.8130E-01,  -3.7434E-01, -3.7672E-01, -3.7935E-01,
+     -3.8230E-01, -3.8563E-01, -3.8945E-01, -3.9390E-01, -3.9919E-01, -4.0563E-01, -4.1376E-01,
+     -4.2448E-01, -4.3965E-01, -4.6363E-01, -5.1055E-01, -6.8316E-01, 6.8316E-01,  5.1055E-0
```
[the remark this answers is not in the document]
SAT-oO: oops forgot to delete that after copying Finns work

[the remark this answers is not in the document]
SAT-oO: I read the same page too after the conversation lol not it should be good

src/application/controller/gain_scheduling.c
```diff
@@ -1 +1,3471 @@
...
+     3.9113E-01,  3.8864E-01,  -3.8130E-01, -3.8373E-01, -3.8643E-01, -3.8945E-01, -3.9286E-01,
+     -3.9678E-01, -4.0134E-01, -4.0676E-01, -4.1337E-01, -4.2170E-01, -4.3271E-01, -4.4828E-01,
+     -4.7290E-01, -5.2112E-01, -6.9870E-01, 6.9870E-01,  5.2112E-01,  4.7290E-01,  4.4828E-01,
+     4.3271E-01,  4.2170E-01,  4.1337E-01,  4.0676E-01,  4.0134E-01,  3.9678E-01,  3.9286E-01,
+     3.8945E-01,  3.8643E-01,  3.8373E-01,  3.8130E-01,  -3.7434E-01, -3.7672E-01, -3.7935E-01,
+     -3.8230E-01, -3.8563E-01, -3.8945E-01, -3.9390E-01, -3.9919E-01, -4.0563E-01, -4.1376E-01,
+     -4.2448E-01, -4.3965E-01, -4.6363E-01, -5.1055E-01, -6.8316E-01, 6.8316E-01,  5.1055E-01,
+     4.6363E-01,  4.3965E-01,  4.2448E-01,  4.1376E-01,  4.0563E-01,  3.9919E-01,  3.9390E-01,
+     3.8945E-01,  3.8563E-01,  3.8230E-01,  3.7935E-01,  3.7672E-01,  3.7434E-01,  -3.6773E-01,
+     -3.7006E-01, -3.7263E-01, -3.7551E-01, -3.7877E-01, -3.8250E-01, -3.8684E-01, -3.92
```
Joe-Joe-Joe-Joe, line 3464: Why are these parameters pointers? And following project convention, this should probably take a pointer to the gain as a param and return a w_status_t. For example, if `arm_bilinear_interp_f32` returns 0 (presumably out of bounds, unless any of our gains are actually 0?) that should probably be an error

Joe-Joe-Joe-Joe, line 3454:
```suggestion
static arm_bilinear_interp_instance_f32 *gain_instance_arr[GAIN_NUM] = {
{ .numRows = GAIN_P_SIZE, .numCols = GAIN_C_SIZE, .pData = &gain_table[0][0] },
{ .numRows = GAIN_P_SIZE, .numCols = GAIN_C_SIZE, .pData = &gain_table[1][0] },
{ .numRows = GAIN_P_SIZE, .numCols = GAIN_C_SIZE, .pData = &gain_table[2][0] },
{ .numRows = GAIN_P_SIZE, .numCols = GAIN_C_SIZE, .pData = &gain_table[3][0] },
};
```
I suppose this defeats the purpose of having `GAIN_NUM` be dynamic, but I would like to avoid adding functions where possible

src/application/controller/controller.c
```diff
@@ -1,28 +1,46 @@
...
 #define CONTROLLER_CYCLE_TIMEOUT_MS 5
 
 static controller_t controller_state = {0};
-static controller_input_t controller_input __attribute__((unused)) = {0};
 static controller_output_t controller_output = {0};
-static controller_gain_t controller_gain __attribute__((unused)) = {0};
+static controller_gain_t controller_gain = {0};
+
+// output related const
+static const float max_commanded_angle = 20 * 180.0 / M_PI; // 20 degrees in radians
```
Joe-Joe-Joe-Joe, line 20: You need to flip your 180 and M_PI. Same with the CAN send function

Joe-Joe-Joe-Joe (changes requested):
Few nits, looking very good overall

Joe-Joe-Joe-Joe:
Same issue as before, the inline comments do not render properly on this page 💀

src/application/controller/controller.c
SAT-oO: 💀 im trolling so f badly

src/application/controller/gain_scheduling.c
SAT-oO: I had this at first lol then changed it lol

SAT-oO: 🤔 I dunno how much a 0 in gain will affect the canard cmd cuz we're taking the dot product after this. I'll ask in channel for clarification

CMakeCache.txt
celery6, line 1: why so many extraneous cmake files.. be careful when u presscommit

src/application/controller/controller.c
```diff
@@ -1,28 +1,46 @@
...
 
-/*
-    TODO Send `canard_angle`, the desired canard angle (radians) to CAN
-*/
+// Send `canard_angle`, the desired canard angle (radians) to CAN
 static w_status_t controller_send_can(float canard_angle) {
-    // Build the CAN msg using [canard-specific canlib function to be defined later].
+    // convert canard angle from radians to millidegrees
+    int16_t canard_cmd_signed = (int16_t)(canard_angle * M_PI / 180.0 * 1000.0);
+    uint16_t canard_cmd = (uint16_t)canard_cmd_signed;
```
celery6, line 27: are u sure this cast works how u want? idt C treats this like abs value, and idk what this conversion is supposed to do in this situation

src/application/controller/controller.c
```diff
@@ -1,28 +1,46 @@
...
+    float time_ms;
+    if (W_SUCCESS != timer_get_ms(&time_ms)) {
+        time_ms = 0.0f;
+    }
+    uint32_t can_timestamp = (uint32_t)time_ms;
+
+    // Build the CAN msg
+    can_msg_t msg;
+    build_actuator_analog_cmd_msg(
+        PRIO_HIGH, can_timestamp, ACTUATOR_CANARD_ANGLE, canard_cmd, &msg
```
celery6, line 39: change priority to PRIO_HIGHEST as per https://waterloorocketry.slack.com/archives/C07MX0QDS/p1742859436346659?thread_ts=1742859323.024799&cid=C07MX0QDS

src/application/controller/controller.c
Joe-Joe-Joe-Joe: We absolutely do not want to abs this value, rocket will unalive itself if we do

MCB needs to be able to unpack this value as a signed integer to know which way to move the canards

SAT-oO: just wrote a sample program and it works pretty much as expected
when the value is negative (so like -20 for example) the uint16_t cast returns [max range] +value +1 (so 65516)

Joe-Joe-Joe-Joe: I also tested this and was also satisfied the conversion to/from unsigned works as expected

src/application/controller/gain_scheduling.c
```diff
@@ -3451,21 +3452,28 @@ static const float gain_table[GAIN_NUM][GAIN_P_SIZE * GAIN_C_SIZE] = {
...
-float interpolate_gain(float *p_dyn, float *coeff, int gain_index) {
+w_status_t interpolate_gain(float p_dyn, float coeff, controller_gain_t *gain_output) {
     // Normalize coordinates
-    float p_norm = (*p_dyn - pressure_dynamic_offset) / pressure_dynamic_scale - 1;
-    float c_norm = (*coeff - canard_coeff_offset) / canard_coeff_scale - 1;
+    float p_norm = (p_dyn - pressure_dynamic_offset) / pressure_dynamic_scale - 1;
+    float c_norm = (coeff - canard_coeff_offset) / canard_coeff_scale - 1;
+
+    // check bounds for p and c
+    if (p_norm < MIN_COOR_BOUND || p_norm > GAIN_P_SIZE - 1 || c_norm < MIN_COOR_BOUND ||
```
Joe-Joe-Joe-Joe, line 3468: Operator precedence - I don't remember what the rules are, but best practice is to wrap your operations in parentheses so nobody has to

Joe-Joe-Joe-Joe (changes requested):
Please delete the other files you added

src/application/controller/gain_scheduling.c
```diff
@@ -1 +1,3479 @@
...
+#define MIN_COOR_BOUND 0
+
+static const float pressure_dynamic_scale = 3.0619E+03;
+static const float canard_coeff_scale = 6.6667E-01;
+
+static const float pressure_dynamic_offset = 1.0000E+02;
+static const float canard_coeff_offset = -1.0000E+01;
+
+// Gain array
+static const float gain_table[GAIN_NUM][GAIN_P_SIZE * GAIN_C_SIZE] = {
```
FinnBreu, line 15: Should I adjust the creation script to output this every time?

SAT-oO (commented):
Please review 🆘 ignore the pipeline cuz it's just formatting after I solved merge conflict

FinnBreu (commented):
Even though I get that this works, it's still weird to me to call a function in an I'd statement

src/application/controller/controller_algorithm.h
```diff
@@ -0,0 +1,33 @@
...
+#include "application/controller/gain_table.h"
+#include "arm_math.h"
+#include "third_party/rocketlib/include/common.h"
+#include <math.h>
+
+#define FEEDBACK_GAIN_NUM (GAIN_NUM - 1)
+#define MIN_COOR_BOUND 0
+
+// output related const
+static const float max_commanded_angle = 20 / 180.0 * M_PI; // 20 degrees in radians
```
FinnBreu, line 13: This should be 10 degrees, as it's doing + and - this in the limiter

FinnBreu (commented):
All that I can speak to looks really good now, great job! (will approve this after the max angle is correct)
The gain_table files look good and very easily creatable, thanks for that.

SAT-oO:
> Even though I get that this works, it's still weird to me to call a function in an I'd statement

I think for now its just so we can do error logging when flight conditions are out of bound

FinnBreu (approved):
Great stuff

src/application/controller/controller_algorithm.h
```diff
@@ -0,0 +1,33 @@
...
+#include <math.h>
+
+#define FEEDBACK_GAIN_NUM (GAIN_NUM - 1)
+#define MIN_COOR_BOUND 0
+
+// output related const
+static const float max_commanded_angle = 10 / 180.0 * M_PI; // 10 degrees in radians
+static const float reference_signal = 0.0f; // no roll program for test flight
+
+static const float commanded_angle_zero = 0.0f; // safe mode, init overwrite, p and c out of bound
```
celery6, line 16: having a static var in a .h file isn't good in C (unless u deliberately need it for some reason). This creates a new copy of each static var in memory for every file that #includes this .h file.

just use #define since these are constant expressions.

or: declare the vars (without static) in the .h `const float ref_signal;`, and define them in the .c `const float ref_signal = 0.0f`. Since declarations dont take up memory space, there wont be duplicate copies of the vars

src/application/controller/gain_table.h
```diff
@@ -0,0 +1,19 @@
...
+// for building each interpolation instance
+#define GAIN_P_SIZE 200
+#define GAIN_C_SIZE 30
+
+// gain table variables
+static const float pressure_dynamic_scale = 3.0619E+03;
+static const float canard_coeff_scale = 6.6667E-01;
+
+static const float pressure_dynamic_offset = 1.0000E+02;
+static const float canard_coeff_offset = -1.0000E+01;
```
celery6, line 15: see other comment about static vars in .h file

celery6 (approved):
LGTM!??!!?!?! did not look too closely at the math ngl hope joe and finn looked closer but if it passes tests then it works WOOHOO

Joe-Joe-Joe-Joe (approved):
LGTM 
i need to sleep why does this team operate at 1am :face:

PR #173, opened 2025-03-26
Logger implementation (WIP)

martendo:
PR for preliminary review, more to come in the coming days (see TODO comments)


martendo:
Module design in Google Doc is slightly out of date now, whoops

will fix tomorrow

src/application/logger/log.h
```diff
@@ -37,27 +40,35 @@
...
 } log_data_type_t;
 
 #undef M
 
 /**
  * The container for data to be included in messages from log_data().
+ * Make sure to update format descriptions in scripts/logparse.py too!
  */
 typedef union {
     struct {
```
celery6, line 74: i dont think the `packed` doesnt apply recursive to nested structs in this union? if they all need to be packed then each struct needs the attribute

martendo: I was referencing [GCC's docs on the packed attribute](https://gcc.gnu.org/onlinedocs/gcc-4.7.2/gcc/Type-Attributes.html) and I interpreted it to apply recursively but now that I reread it I think you're right :(

martendo: Addressed in 6477b911b89410581f02a968b75fa724ce91bd9c

celery6 (approved):
LGTM!?!?!?!!!!!

PR #175, opened 2025-03-27
Fix Coverage Build Process

taisirhassan:
Fixed coverage report behaviour + fixed broken build process.

### DIFF
--- a/src/drivers/sd_card/sd_card.h
+++ b/src/drivers/sd_card/sd_card.h
@@ -1,10 +1,10 @@
 #ifndef SD_CARD_H
 #define SD_CARD_H
 
-#include <stdbool.h>
-#include <stdint.h>
 #include "FreeRTOS.h"
 #include "semphr.h"
+#include <stdbool.h>
+#include <stdint.h>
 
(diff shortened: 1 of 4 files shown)


Joe-Joe-Joe-Joe (approved):
do we care

image

otherwise LGTM

celery6 (dismissed):
YAYYYYY WE'RE SO BACK

src/drivers/sd_card/sd_card.h
```diff
@@ -19,6 +21,10 @@ typedef struct {
     uint32_t err_count;
 } sd_card_health_t;
 
+// Declare these as extern to allow access from tests
```
celery6, line 24: please dont do this in the sd_card header cuz this reveals it to every single other file... which is not what we want. do the `extern` inside the test cpp file instead to minimize damage of making it non-static

celery6 (changes requested):

celery6 (approved):
moment moment

taisirhassan:
> moment moment

:face skull

PR #182, opened 2025-04-01
Estimator implementation: Pad Filter

lukeqin-s:
### Pad Filter Implementation:

* Added `pad_filter.c` and `pad_filter.h`: with `pad_filter` function to compute the initial state and determine IMU biases.
* Relevant `.c` and `.h` files from other not yet merged branches have been copied into this branch.


src/application/estimator/pad_filter.c
```diff
@@ -0,0 +1,237 @@
...
+    // set constant initials
+
+    const float alpha = 0.005; // low pass time constant
+
+    const vector3d_t w = {{0.0f}}; // stationary on rail
+    const vector3d_t v = {{0.0f}}; // stationary on rail
+
+    const float Cl = 0;
+
+    const float delta = 0; // controller sets canards to zero due to flight phase
```
SAT-oO, line 23: static, check other variables too

SAT-oO, line 23: declare them on the top of the file and put static, other matlab portions are using this constants too -> will put them all tgt at some point

src/application/estimator/pad_filter.c
```diff
@@ -0,0 +1,237 @@
...
+        data->movella.gyroscope.z,
+        data->movella.magnometer.x,
+        data->movella.magnometer.y,
+        data->movella.magnometer.z,
+        data->movella.barometer
+    };
+
+    // Static filtered data of IMU_i, remembers from last iteration
+    static float filtered_1[10];
+    static float filtered_2[10];
```
SAT-oO, line 72: define a const for this 10 and put it at the top plz, otherwise its just a magic int in the program

src/application/estimator/pad_filter.c
```diff
@@ -1,154 +1,134 @@
...
-    const vector3d_t v = {{0.0f}}; // stationary on rail
-
-    const float Cl = 0;
+static const float low_pass_alpha = 0.005; // low pass time constant
+
+// set constant initials - knowing that the rocket is stationary on the rail
+static const vector3d_t w = {{0.0f}}; // stationary on rail
+static const vector3d_t v = {{0.0f}}; // stationary on rail
+static const float Cl = 5; // estimated coefficient of lift, const with Ma
+static const float delta = 0; // controller sets canards to zero due to flight phase
```
SAT-oO, line 13: we so need a file for all the constants 💀

src/application/estimator/pad_filter.c
```diff
@@ -1,154 +1,134 @@
...
+
+// Computes inital state and covariance estimate for EKF, and bias values for the IMU
+// Uses all available sensors: Gyroscope W, Magnetometer M, Accelerometer A, Barometer P
+// Outputs: initial state, sensor bias matrix, [x_init, bias_1, bias_2]
+w_status_t pad_filter(
+    pad_filter_ctx_t *ctx, const y_imu_t *IMU_1, const y_imu_t *IMU_2, const bool is_dead_1,
+    const bool is_dead_2, x_state_t *x_init, y_imu_t *bias_1, y_imu_t *bias_2
+) {
+    if ((IMU_1 == NULL) || (IMU_2 == NULL) || (x_init == NULL) || (bias_1 == NULL) ||
+        (bias_2 == NULL)) {
```
SAT-oO, line 26: do we care about BARR or is it applicable in here 🤔

src/application/estimator/pad_filter.c
```diff
@@ -159,59 +139,49 @@ w_status_t pad_filter(estimator_all_imus_input_t *data) {
 
     // conconct state vector
 
-    float x_init[13] = {q.w, q.x, q.y, q.z, w.x, w.y, w.z, v.x, v.y, v.z, alt, Cl, delta};
+    float x_init_out[13] = {q.w, q.x, q.y, q.z, w.x, w.y, w.z, v.x, v.y, v.z, alt, Cl, delta};
```
SAT-oO, line 142: why not use the new x_state_t for full state in the estimator_types file? this way you can define the w, v, Cl and whatever in the function instead of up at the top cuz they don't rly represent constants anyways

src/application/estimator/model/quaternion.c
```diff
@@ -37,14 +37,17 @@ matrix3d_t quaternion_rotmatrix(const quaternion_t *q_unnormed) {
...
+    // middle row
+    S.s21 = 2 * (q.x * q.y - q.w * q.z);
     S.s22 = 1 - 2 * (q.x * q.x + q.z * q.z);
-    S.s23 = 2 * (q.y * q.z - q.w * q.x);
-    S.s31 = 2 * (q.x * q.z - q.w * q.y);
-    S.s32 = 2 * (q.y * q.z + q.w * q.x);
+    S.s23 = 2 * (q.y * q.z + q.w * q.x);
+    // bottom row
+    S.s31 = 2 * (q.x * q.z + q.w * q.y); 
+    S.s32 = 2 * (q.y * q.z - q.w * q.x);
```
SAT-oO, line 50: was this bcz genuinely it was inputted wrong or was it just so the unit tests for pad filter pass 💀 
cuz the main diagonal entries wasn't changed which indicates emmmm pad filter issues ... other models use the q_rotmatrix as well so im not sure if the other unit tests stay true once this gets merged

src/application/estimator/pad_filter.c
FinnBreu: yes please do

src/application/estimator/model/quaternion.c
FinnBreu: some off-diagonals were wrong

src/application/estimator/pad_filter.c
celery6: ngl its kind of nicer to have the constants in each model file instead of shoved into 1 massive global constants file. thats so chaotic. but this makes it actually easy to find the constants that are relevant to this model. ive seen projects that overuse smth like "global_constants.h" and glhf scrolling through 800 lines of program-wide constants with no organization!!!

src/application/estimator/pad_filter.c
celery6: yes we care good catch

src/application/estimator/pad_filter.c
celery6: ..are they not constants? they represent the constant state of when the rocket is stationary on the pad. so i left those as static consts at the top of the file instead of hardcoding numbers into the function

SAT-oO: (I don't have context lol) keep the state constants at the top but use x_state_t

celery6 (approved):
thank luke orzorzoroz

src/application/estimator/pad_filter.h
```diff
@@ -0,0 +1,37 @@
...
+#include "common/math/math.h"
+#include "third_party/rocketlib/include/common.h"
+#include <stdbool.h>
+#include <stdint.h>
+
+/**
+ * this holds persistent data for 1 instance of a pad_filter (ie, its context)
+ */
+typedef struct {
+    float filtered_1_arr[10];
```
Joe-Joe-Joe-Joe, line 15: Why don't we use `y_imu_t` here? This represents the same data format, right? Cause then the source file is full of magic number indices

celery6: because all the matlab code is done on the filtered_arr arrays as magic number indices 💀

Joe-Joe-Joe-Joe: Okay but can you change it

celery6: i intend for the C code to align exactly with matlab much as possible because that's the source of truth, so using arr actually improves readability here imo (readable = u can compare between C and Matlab at a glance)

src/application/estimator/pad_filter.c
```diff
@@ -4,13 +4,13 @@
...
+static const double low_pass_alpha = 0.005; // low pass time constant
 
 // set constant initials - knowing that the rocket is stationary on the rail
-static const vector3d_t w = {{0.0f}}; // stationary on rail
-static const vector3d_t v = {{0.0f}}; // stationary on rail
-static const float Cl = 5; // estimated coefficient of lift, const with Ma
-static const float delta = 0; // controller sets canards to zero due to flight phase
+static const vector3d_t w = {{0.0}}; // stationary on rail
+static const vector3d_t v = {{0.0}}; // stationary on rail
+static const double Cl = 5; // estimated coefficient of lift, const with Ma
```
SAT-oO, line 12: thought we're staying consistent with matlab const names?
either all those goes into const spreadsheet or define cl_alpha at the top then assign the CL to equal to cl_alpha 
low_pass_alpha needs to go into spreadsheet for sure

src/application/estimator/pad_filter.c
```diff
@@ -46,20 +39,21 @@ w_status_t pad_filter(
     }
 
     // Initialization - only run 1 time in the whole program
+    // TODO: could optimize init to be done once outside of pad_filter?
```
SAT-oO, line 42: is it possible we move this to a separate init function, cuz ekf also needs to init, so as things run just under estimator init we init all its submodules like here in pad filter and in ekf

SAT-oO: this is also a lil confusing structure wise when comparing between matlab code and C

src/application/estimator/pad_filter.c
```diff
@@ -87,59 +81,70 @@ w_status_t pad_filter(
...
+        a.z += ctx->filtered_2.accelerometer.z;
     }
 
     // Normalize the acceleration by the number of alive IMUs
-    a[0] /= (float)num_alive_imus;
-    a[1] /= (float)num_alive_imus;
-    a[2] /= (float)num_alive_imus;
+    a.x /= (double)num_alive_imus;
+    a.y /= (double)num_alive_imus;
+    a.z /= (double)num_alive_imus;
```
SAT-oO, line 101: we have helper functions called vector add and vector scale? given they are of type vector3d_t now instead of arrays of length 3

celery6: omg those exist orz

src/application/estimator/pad_filter.c
```diff
@@ -87,59 +81,70 @@ w_status_t pad_filter(
...
+    x_init->attitude.w = q.w;
+    x_init->attitude.x = q.x;
+    x_init->attitude.y = q.y;
+    x_init->attitude.z = q.z;
+    x_init->rates.x = w.x;
+    x_init->rates.y = w.y;
+    x_init->rates.z = w.z;
+    x_init->velocity.x = v.x;
+    x_init->velocity.y = v.y;
+    x_init->velocity.z = v.z;
```
SAT-oO, line 148: do we care about aesthetics and be like `x_init->attitude = q;`

tests/unit/pad_filter_test.cpp
```diff
@@ -430,7 +352,7 @@ TEST(PadFilterTest, RunsThreeSequentialIterationsCorrectly) {
         );
 
         for (int i = 0; i < 13; ++i) {
-            EXPECT_NEAR(actual_x.array[i], expect_x.array[i], fabs(expect_x.array[i] * tolerance));
+            EXPECT_NEAR(actual_x.array[i], expect_x.array[i], abs(expect_x.array[i] * tolerance));
         }
         for (int i = 0; i < 10; ++i) {
             EXPECT_NEAR(
```
SAT-oO, line 362: change the fabs below to abs

src/application/estimator/pad_filter.c
celery6: orz i forgor u can just assign structs to each other

src/application/estimator/pad_filter.c
celery6: this is copypasted from matlab though? matlab contains this init thing. the TODO is a proposal to deviate from matlab when we have time to do that

tests/unit/pad_filter_test.cpp
celery6: yup so true

src/application/estimator/pad_filter.c
SAT-oO: yea but the copying does always make sense in c
like in ekf the arm_mat_init stuff has to be placed in a function, so there has to be a separate init for all the matrix instances, im assuming putting everything here makes it look cleaner + less prone to mistakes when those doesn't get mixed up with the things that iterates every 5ms

tests/unit/pad_filter_test.cpp
```diff
@@ -110,11 +119,28 @@ TEST(PadFilterTest, NewContextOneIteration) {
...
+    ASSERT_EQ(pad_filter_init(&test_ctx, &imu1, &imu2, is_dead_1, is_dead_2), W_FAILURE);
+}
+
 TEST(PadFilterTest, RunsThreeSequentialIterationsCorrectly) {
     // Arrange
     pad_filter_ctx_t test_ctx = {0};
-    // tolerance is evaled as 0.1% of the given expect value
-    float tolerance = 0.005;
+    // tolerance is evaled as 0.2% of the given expect value
+    float tolerance = 0.002;
```
SAT-oO, line 143: lol this is so real 💀

SAT-oO (approved):
looks gooddd

tests/unit/pad_filter_test.cpp
celery6: it's improvement!!!!

celery6: 3 iterations seems to kill the precision.. sooo idk what to do. the code is correct so we can evaluate and fix precision later 😭

src/application/estimator/pad_filter.c
celery6: it's attached to a commit in Matlab, imo that's way more clear and helpful than the spreadsheet for this 😭

SAT-oO: I mean yea im fine with this but isn't the whole point to have C as sth readable on its own without needing matlab to much 🤔

celery6: .. is that the point now? i still thinkt the point of this is to copy matlab  💀

celery6: I don't rly understand the purpose of the spreadsheet for the estimator implementation code ngl. cuz everything in the spreadsheet is just be a pointer to some matlab commit. That seems like the sheet is an arbitrary layer of indirection  @Joe-Joe-Joe-Joe

SAT-oO: while maintaining certain readability standards plus aspects easier for checking?

SAT-oO: whatever later issue

PR #185, opened 2025-04-01
Controller unit test

SAT-oO:
tolerance for both cmd angle and gain values set to .001 of the expected values
stays failed for now as more tuning is done 
tests/gain table/offset and scaling constants will be updated as things go


celery6 (commented):
merge main into ur branch? why does this show so many changes that arent actually changed

celery6 (approved):

PR #187, opened 2025-04-02
Estimator models

FinnBreu:


celery6:
dynamics model in #189  and imu model in #195

PR #190, opened 2025-04-04
Orientation correction

tri723:
It is matrix multiplication by correction matrices used over and over again. It is working, only BARR standards may have to be updated.


src/application/imu_handler/imu_handler.c
```diff
@@ -17,6 +17,14 @@
 #define ACCEL_FRESHNESS_TIMEOUT_MS 5
 #define BARO_FRESHNESS_TIMEOUT_MS 25
 
+// Update mats correct orientation
+static const matrix3d_t g_movella_upd_mat = {.array = {{1, 0, 0}, {0, 1, 0}, {0, 0, 1}}};
```
Joe-Joe-Joe-Joe, line 21: please make sure these are added to the spreadsheet: https://docs.google.com/spreadsheets/d/1s6TOQ8qk1PcYXq2-ggQLLja72tSYZh3B149Vd1C0oUw/edit?gid=0#gid=0

src/application/imu_handler/imu_handler.c
```diff
@@ -56,6 +64,14 @@ static w_status_t initialize_all_imus(void) {
     return status;
 }
 
+static vector3d_t apply_orientation_correction(const vector3d_t *vec, const matrix3d_t *mat){
+    g_inp_vec.x = vec->x;
+    g_inp_vec.y = vec->y;
+    g_inp_vec.z = vec->z;
+
+    return math_vector3d_rotate(mat, &g_inp_vec);
```
Joe-Joe-Joe-Joe, line 72: `math_vector3d_rotate` takes `const vector3d_t *vector`, which means you can safely pass the pointer without an intermediary variable. I would invert the parameter order for clarity

src/application/imu_handler/imu_handler.c
tri723: Got it 👍

src/application/imu_handler/imu_handler.c
tri723: Completely missed that, the code is much cleaner. Done and done.

src/application/imu_handler/imu_handler.c
```diff
@@ -74,6 +78,11 @@ static w_status_t read_pololu_imu(estimator_imu_measurement_t *imu_data) {
     status |= altimu_get_baro_data(&baro_data);
 
     if (W_SUCCESS == status) {
+        // Applies orientation correction
+        imu_data->accelerometer = math_vector3d_rotate(&(imu_data->accelerometer), &g_polulu_upd_mat);        
+        imu_data->gyroscope = math_vector3d_rotate(&(imu_data->gyroscope), &g_polulu_upd_mat);
+        imu_data->magnometer = math_vector3d_rotate(&(imu_data->magnometer), &g_polulu_upd_mat);
```
FinnBreu, line 84: Check the function definition, the argument order is matrix first, vector second

src/common/math/math-algebra3d.c
```diff
@@ -55,7 +55,7 @@ vector3d_t math_vector3d_cross(const vector3d_t *vector_left, const vector3d_t *
 }
 
 // matrix * vector // matrix vector multiplication, for rotating vectors
-vector3d_t math_vector3d_rotate(const matrix3d_t *matrix, const vector3d_t *vector)
+vector3d_t math_vector3d_rotate(const vector3d_t *vector, const matrix3d_t *matrix)
```
FinnBreu, line 58: You cannot just change the function definition of a shared library! Change the order in your code

src/common/math/math-algebra3d.h
```diff
@@ -24,7 +24,7 @@ float math_vector3d_dot(const vector3d_t *vector1, const vector3d_t *vector2);
 vector3d_t math_vector3d_cross(const vector3d_t *vector_left, const vector3d_t *vector_right);
 
 // matrix * vector // matrix vector multiplication, for rotating vectors
-vector3d_t math_vector3d_rotate(const matrix3d_t *matrix, const vector3d_t *vector);
+vector3d_t math_vector3d_rotate(const vector3d_t *vector, const matrix3d_t *matrix);
```
FinnBreu, line 27: See above

FinnBreu (changes requested):
Correct the argument order in your code, instead of changing approved common libraries

src/common/math/math-algebra3d.c
tri723: I should have known, won't do this recklessly again!

src/application/imu_handler/imu_handler.c
FinnBreu: only this is missing now?

Joe-Joe-Joe-Joe: I have added it. Will need to consult Ben for final orientation in CAD

celery6: can we merge this with the identity matrix and update numbers later

Joe-Joe-Joe-Joe: Yes

tests/unit/imu_handler_test.cpp
```diff
@@ -105,7 +127,7 @@ class ImuHandlerTest : public ::testing::Test {
         RESET_FAKE(xTaskGetTickCount);
         RESET_FAKE(xTaskDelayUntil);
```
SAT-oO, line 128: Where is this line used 🤔 
Plus those are't fake defined

tests/unit/imu_handler_test.cpp
```diff
@@ -214,18 +218,14 @@ TEST_F(ImuHandlerTest, RunWithPoluluFailure) {
...
 
     // Verify Movella data is still correct and not dead
-    EXPECT_FLOAT_EQ(EXPECTED_ACC.x, captured_data.movella.accelerometer.x);
-    EXPECT_FLOAT_EQ(EXPECTED_GYRO.x, captured_data.movella.gyroscope.x);
-    EXPECT_FLOAT_EQ(EXPECTED_MAG.x, captured_data.movella.magnetometer.x);
-    EXPECT_FLOAT_EQ(EXPECTED_BARO, captured_data.movella.barometer);
+    EXPECT_TRUE(vectors_are_equal(captured_data.movella.accelerometer, EXPECTED_ACC, tolerance));
+    EXPECT_TRUE(vectors_are_equal(captured_data.movella.gyroscope, EXPECTED_GYRO, tolerance));
+    EXPECT_TRUE(vectors_are_equal(captured_data.movella.magnetometer, EXPECTED_MAG, tolerance));
+    EXPECT_NEAR(captured_data.movella.barometer, EXPECTED_BARO, abs(EXPECTED_BARO * tolerance));
```
SAT-oO, line 228: This tolerance is a lil high for baro

SAT-oO (changes requested):
Line 117 is a lil redundant too given you've already initialized that at complie time

tests/unit/imu_handler_test.cpp
celery6: whatt its literally 0.005%??

SAT-oO: yea true but that was not a 6-7 digit overlap 
if it doesn't need to get that precise then ignore the comment

celery6: its 5 digit overlap
doesnt this method of doing tolerance as a % of expected val mean every expect_near here is within 5 digits? so if baro isn't accurate then nothing else is either :face:

SAT-oO: ya ik its a 5 digit overlap 
dunno if that's good enough for the standard we are going doesn't hurt to bring it up 👐

tests/unit/imu_handler_test.cpp
celery6: these r fake defined in the mocks folder
deleted taskgettickcount but the other is actually used in imu_handler.c

tests/unit/imu_handler_test.cpp
celery6: i think we've already executive-decisioned that opimtizing precision is a later thing :dies 
prioritize make  code exist first :skull:

tests/unit/imu_handler_test.cpp
SAT-oO: Where xtaskdelayuntil 🤔 
There's only vtaskdelayuntil

celery6: skull i bet someone chatgptd too hard earlier without looking..

SAT-oO: I swear im not blind 😢

SAT-oO (approved):
Looks goodd

PR #200, opened 2025-04-06
Add CAN logging for state estimation in estimator module

taisirhassan:
This commit introduces a new function, `estimator_log_state_to_can`, which sends state estimation data over CAN. It iterates through defined state IDs, builds CAN messages, and transmits them. A dummy state is also populated for testing purposes in the estimator's run loop. The timer driver is included to fetch the current time for message timestamps.


src/application/estimator/estimator.c
```diff
@@ -57,6 +59,45 @@ static w_status_t estimator_run_pad_filter(void) {
...
+    }
+    uint16_t timestamp_16bit = (uint16_t)current_time_ms;
+
+    // Iterate through all defined state IDs
+    for (can_state_est_id_t state_id = 0; state_id < STATE_ID_ENUM_MAX; ++state_id) {
+        // The x_state_t union maps directly to the enum order if accessed as an array
+        if (!build_state_est_data_msg(
+                PRIO_LOW, timestamp_16bit, state_id, &current_state->array[state_id], &msg
+            )) {
+            // log_text("Estimator", "Failed to build CAN message for state ID %d", state_id);
```
celery6, line 87: uncomment the log??? logger exists now

tests/unit/estimator_test.cpp
celery6, line 1: please add test case(s) to ensure ur feature works. espeically that its sending the right values, and only every 5 cycles

taisirhassan:
> Code LGTM, test seems comprehensive, def needs a test IRL. Omnibus/parsley should be up to date with state format so this should just work

pog @celery6 thoughts?

celery6 (changes requested):
https://github.com/waterloo-rocketry/cansw_processor_canards/commit/754b23cf40b3ce23b131a01725f748099b387bde

skull emoji. the downsides of joe being so scared of ifdefs that we're doing this sus method of run_loop functions that miss the actual freertos task loop... skill issue but whatever its fine for now

Joe-Joe-Joe-Joe:
I don't think my fear is unjustified. We have created silent hardware failure via changes to build scripts before. I'll admit I don't exactly understand your issue here, but I think it is always worth putting as much of the "awkward code" burden on the unit test side, as opposed to the source code side (such as using extern over conditional header declarations)

celery6:
of all things, having `#ifdef UNIT_TEST_BUILD` is the most industry standard thing possible. This method avoids affecting the flight code, as the preprocessor processes it out.

Meanwhile, for example, removing a `static` to cater to unit tests actually changes the flight code as it will compile without a `static` where there should be a `static`.

Isn't that literally what ur agreeing with?
>  I think it is always worth putting as much of the "awkward code" burden on the unit test side, as opposed to the source code side (such as using extern over conditional header declarations)

Put the burden on code inside ifdef guards that will never touch the source code. If you don't trust this preprocessor macro working, u should raise an issue with all the other macros we have, including the mandatory CAN defines, stm32 defines, NDEBUG defines, etc...

anyway it's too late to change this so this is just a rant with no action

celery6 (approved):

celery6:
approved by taisir

PR #202, opened 2025-04-07
Test math libs

celery6:


src/application/estimator/model/quaternion.c
```diff
@@ -0,0 +1,120 @@
...
+//     euler.y = atan2f(
+//         2.0f * (q_norm.y * q_norm.z + q_norm.w * q_norm.x),
+//         q_norm.w * q_norm.w - q_norm.x * q_norm.x - q_norm.y * q_norm.y + q_norm.z * q_norm.z
+//     );
+
+//     return euler;
+// }
+
+// // Compute Euler angle roll from a quaternion
+// float quaternion_to_roll(const quaternion_t *q) {
```
SAT-oO, line 114: we might need to Euler stuff at some point but let's just leave it for now 😅

tests/unit/common_math_test.cpp
```diff
@@ -0,0 +1,195 @@
...
+
+// Test case: Non-symmetric matrix should transpose into a different matrix
+TEST(CommonMathTest, NonSymmetricMatrix) {
+    matrix3d_t input = {1.0f, 2.0f, 3.0f, 4.0f, 5.0f, 6.0f, 7.0f, 8.0f, 9.0f};
+
+    matrix3d_t expected = {1.0f, 4.0f, 7.0f, 2.0f, 5.0f, 8.0f, 3.0f, 6.0f, 9.0f};
+
+    matrix3d_t result = math_matrix3d_transp(&input);
+
+    EXPECT_TRUE(matrices_are_equal(result, expected));
```
SAT-oO, line 86: isn't this the same as the basic test?

SAT-oO (changes requested):
I didn't do the math for rotmatrix/derivative, but hey if it works it works

tests/unit/quaternion_test.cpp
```diff
@@ -118,15 +118,61 @@ TEST(QuaternionTest, KnownQuaternionTestNormalize) {
...
-        0.0f, 0.0f, 1.0f, 0.0f, 1.0f, 0.0f, -1.0f, 0.0f, 0.0f
-    }; // Expected rotation matrix
-
-    matrix3d_t result_matrix = quaternion_rotmatrix(&q);
+    float tolerance = 1e-3f;
 
-    EXPECT_TRUE(matrices_are_equal(result_matrix, expected_matrix));
+    // --- Test Case 1: 90 deg yaw ---
+    quaternion_t q1 = {0.7071f, 0.0f, 0.0f, 0.7071f}; // 90 deg yaw
+    // MATLAB output (column-major): [0 -1 0; 1 0 0; 0 0 1]
```
SAT-oO, line 125: this format of an output is row major? 

if M is a matrix `disp(M)` gives row major, `disp(M(:))` gives column major, transpose then linearize gives row major too

src/application/estimator/model/quaternion.c
```diff
@@ -37,14 +37,17 @@ matrix3d_t quaternion_rotmatrix(const quaternion_t *q_unnormed) {
...
+    // middle row
+    S.s21 = 2 * (q.x * q.y - q.w * q.z);
     S.s22 = 1 - 2 * (q.x * q.x + q.z * q.z);
-    S.s23 = 2 * (q.y * q.z - q.w * q.x);
-    S.s31 = 2 * (q.x * q.z - q.w * q.y);
-    S.s32 = 2 * (q.y * q.z + q.w * q.x);
+    S.s23 = 2 * (q.y * q.z + q.w * q.x);
+    // bottom row
+    S.s31 = 2 * (q.x * q.z + q.w * q.y);
+    S.s32 = 2 * (q.y * q.z - q.w * q.x);
```
SAT-oO, line 50: the operation to give matrix transposed as a result

SAT-oO (changes requested):
the entire rotmatrix unit test compares the transpose of expected matrix and actual matrix, mathematically works but actual implementation it might be confusing

- suppose a quaternion [q], its corresponding matrix [M], the transpose of the matrix [M.'], 
- what we have rn is that we input a [q], `rotmatrix` in C outputs [M.']; we take the result from matlab in the row major format of [M] = [a b c; d e f; g h i], manually transpose into [M.'], and compare if the two [M.'] match
- what we want is input [q], output [M] in C, output [M] in matlab, compare entries 

I suggest we change the quaternion rotmatrix back to original or to however it matches the math formula, then get rid of the manual transpose in unit tests

celery6:
approved

PR #206, opened 2025-04-11
Estimator cmake update and use double

celery6:


src/application/estimator/estimator_types.h
```diff
@@ -37,12 +42,12 @@ typedef struct {
 // EKF: bias_i, y_i, and h_x_i are of this type
 // IMU measurement model: the return is of this type
 typedef union {
-    float array[10];
+    double array[Y_IMU_SIZE_FLOATS];
     struct {
         vector3d_t accelerometer;
         vector3d_t gyroscope;
         vector3d_t magnetometer;
```
SAT-oO, line 49: change the quaternion and vector3d_t def to also be double?

celery6: :FACE: i forgot that whole file exists

SAT-oO (approved):
mergeeeee yayyyy

PR #207, opened 2025-04-12
jacobian for airdata plus completed unit test

SAT-oO:


src/application/estimator/model/model_airdata.c
```diff
@@ -69,16 +69,54 @@ estimator_airdata_t model_airdata(double altitude) {
...
+    result.density = result.pressure / (air_R * result.temperature);
 
     // local speed of sound
-    result.mach_local = sqrt(AIR_GAMMA * AIR_R * result.temperature);
+    result.mach_local = sqrt(air_gamma * air_R * result.temperature);
 
     return result;
+}
+
+double model_airdata_jacobian(double altitude) {
```
celery6, line 86: we should try to consistently add a comment to label the C code with the corresponding matlab commit # u were on

src/application/estimator/model/model_airdata.h
```diff
@@ -18,6 +18,14 @@ double model_altdata(double pressure);
 // airdata function uses altitude to return pressure, temperature, density, local mach
 estimator_airdata_t model_airdata(double altitude);
 
+/*
+ * Jacobian of the airdata function
+ * @param altitude 
+ * @return pressure_altitude
```
Joe-Joe-Joe-Joe, line 24: I do not understand what this function returns. Especially considering the jacobian is, in general, a matrix, and this returns a double

Joe-Joe-Joe-Joe: If it returns pressure, calling it "pressure_altitude" is extremely confusing

Joe-Joe-Joe-Joe: Also, I assume altitude is AGL, but in this context you should make it explicit. That's what the docstring is for

tests/unit/airdata_jacobian_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+#define TOLERANCE 0.000001 // non-ratio, checks 6 decimals points
+
+DEFINE_FFF_GLOBALS; // this must be called within the extern c block
+}
+
+class AirdataJacobianTest : public ::testing::Test {
+protected:
+    void SetUp() override {
+        // Reset all fakes before each test, for example:
+        // RESET_FAKE(xQueueCreate);
```
Joe-Joe-Joe-Joe, line 18: Please remove the example comments from the test format when you copy it :face:

Joe-Joe-Joe-Joe (changes requested):

src/application/estimator/model/model_airdata.h
SAT-oO: ig this is where I should break the black box 👿

FinnBreu: It is indeed the pressure change with altitude.
My PDE Prof used this notation where the partial derivative of a variable (function) x with respect to another variable y is x_y.
If this is confusing we could call it x_partial_y, but this gets long fast.

FinnBreu: And I call all of the partial derivatives Jacobian, given that the rest of the matrix is either trivial anyway (eg pressure partial derivatives after all other states is zero), or because it is a sub-matrix of the bigger Jacobian matrix - so the Jacobian of sub-function

tests/CMakeLists.txt
```diff
@@ -311,4 +310,13 @@ add_test_group(test_health_check
     ${UNIT_TESTS_DIR}/health_checks_test.cpp
     # src file(s)
     ${PROC_SRC_DIR}/application/health_checks/health_checks.c
-)
\ No newline at end of file
+)
+
+add_test_group(test_airdata_jacobian
```
Joe-Joe-Joe-Joe, line 315: I feel like this should be rolled into the existing airdata tests, rather than making a new suite

Joe-Joe-Joe-Joe (dismissed):
No functional issues I'm approving this anyway

src/application/estimator/model/model_airdata.c
```diff
@@ -69,16 +70,54 @@ estimator_airdata_t model_airdata(double altitude) {
...
+    // static pressure, different barometric formulas for lapse rate / no lapse rate
+    if (k == 0) {
+        pressure_altitude = -P_B * earth_g0 / (T_B * air_R) * (altitude_ratio * altitude_ratio) *
+                            exp(-earth_g0 * (altitude - b) / (T_B * air_R));
+    } else {
+        pressure_altitude = -P_B * earth_g0 / (T_B * air_R) * (altitude_ratio * altitude_ratio) *
+                            pow(1 - k / T_B * (altitude - b), earth_g0 / (air_R * k) - 1);
+    }
+
+    return pressure_altitude;
```
FinnBreu, line 122: In matlab this also returns the variations of density and mach with altitude. However these are not needed anyway (for testflight at least), so this is good as is.

FinnBreu (dismissed):

tests/unit/airdata_jacobian_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+    void TearDown() override {}
+};
+
+TEST_F(AirdataJacobianTest, NominalCheck) {
+    // Arrange
+    double expected_output[10] = {
+        -8.21305872243281,
+        -4.94526529701412,
+        -4.62642002367724,
+        -5.06763960026881,
```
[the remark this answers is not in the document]
SAT-oO: I manually changed the g to be 9.81 instead of 9.8 (in accordance with the other ones)

tests/unit/model_airdata_test.cpp
```diff
@@ -3,21 +3,30 @@
 
 extern "C" {
 #include "application/estimator/model/model_airdata.h"
-}
-
 // 0.005% tolerance, proportional to the actual value
 #define tolerance 0.00005
+#define TOLERANCE 0.000001 // non-ratio, checks 6 decimals points
```
Joe-Joe-Joe-Joe, line 8: > #define tolerance
> #define TOLERANCE


cmon...

Joe-Joe-Joe-Joe: The fact that you went through the effort to write verbose comments and not to rename them is... sure something

Joe-Joe-Joe-Joe (approved):
I love math

SAT-oO:
everything passes locally pipeline skill issuing 💀 
@celery6 help

src/application/estimator/model/model_airdata.c
```diff
@@ -2,10 +2,10 @@
...
 
 // physical constants
-static const double AIR_R = 287.0579; // specific gas constant
-static const double AIR_GAMMA = 1.4; // adiabatic index for air
-static const double EARTH_R0 = 6356766.0; // mean earth radius
-static const double EARTH_G0 = 9.81; // gravitational acceleration
+static const double air_R = 287.0579; // specific gas constant
+static const double air_gamma = 1.4; // adiabatic index for air
+static const double earth_r0 = 6356766.0; // mean earth radius
+static const double earth_g0 = 9.81; // gravitational acceleration
```
celery6, line 8: @FinnBreu can u change the matlab code to also use 9.81? this is a discrepency? some of the matlab still uses 9.8

celery6 (approved):

PR #215, opened 2025-04-18
Things to merge into main from integration testing

celery6:
changes from integration testing last weekend (april 12)


SAT-oO (approved):
dummy state oopsies

Issue #210, opened 2025-04-13
filter msgs inside CAN RX isr


celery6:
only process the msg types that we care about so the CAN RX freertos task doesn't get flooded
#2

celery6:
the task seems fine and taking not too much time

PR #216, opened 2025-04-18
add msg type filtering in can isr

celery6:
need to test irl
#210


celery6:
implemented by taisir already orz

PR #227, opened 2025-04-28
Imu jacobian

SAT-oO:
created jacobian helper files


src/application/estimator/model/model_imu.c
```diff
@@ -1,11 +1,15 @@
...
 #include "application/estimator/model/model_imu.h"
 #include "application/estimator/estimator_types.h"
+#include "application/estimator/model/jacobians.h"
 #include "application/estimator/model/model_airdata.h"
 #include "application/estimator/model/quaternion.h"
 #include "common/math/math-algebra3d.h"
 #include "common/math/math.h"
 
+// jacobian matrix data struct init
+measurement_model_jacobian_t J = {0};
```
celery6, line 11: using a global variable like this is almost certainly not the best solution for this problem...

src/application/estimator/model/model_imu.c
```diff
@@ -29,3 +33,27 @@ y_imu_t model_measurement_imu(const x_state_t *state, const y_imu_t *imu_bias) {
     return measurement_prediction;
 }
 
+// jacobian of the measurement model
+void model_measurement_imu_jacobian(
+    arm_matrix_instance_f64 *imu_jacobian, const x_state_t *state, const y_imu_t *imu_bias
```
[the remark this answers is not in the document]
celery6: ie, just write directly to `imu_jacobian` instead of writing to some intermediate then copying into `imu_jacobian` at the end???

src/application/estimator/model/jacobians.h
```diff
@@ -0,0 +1,79 @@
+#ifndef JACOBIANS_H
+#define JACOBIANS_H
+
+// JACOBIANS HELPER FUNCTIONS
```
celery6, line 4: are these generic for all jacobians or can they go into the specific jacobian header files? like `measurement_model_jacobian_t` should go into model_imu.h?

src/application/estimator/model/model_imu.c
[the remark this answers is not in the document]
SAT-oO: I did that in ekf forgot to change things 😢

src/application/estimator/model/jacobians.h
SAT-oO: `measurement_model_jacobian_t`is deleted cuz it is no longer needed
all the other ones are the ones needed for the quaternion helper functions in `quaternion.h`

celery6 (approved):

PR #240, opened 2025-05-04
Log fixes

celery6:


src/application/imu_handler/imu_handler.c
```diff
@@ -171,6 +197,11 @@ static w_status_t read_movella_imu(estimator_imu_measurement_t *imu_data) {
     status = movella_get_data(&movella_data, 100); // Add 100ms timeout
 
     if (W_SUCCESS == status) {
+        // convert m/s^2 to gravities
+        movella_data.acc.x = movella_data.acc.x / 9.81f;
```
FinnBreu, line 201: Incorrect, the required unit in the EKF is m/s²

src/application/imu_handler/imu_handler.c
```diff
@@ -27,11 +27,22 @@
...
+static const matrix3d_t g_pololu_upd_mat = {
+    .array =
+        {{0, 0, -1.00000000},
+         {-1.00000000000, 0, 0},
+         {
+             0,
+             1.00000000000,
+             0,
+         }}
+};
```
FinnBreu, line 45: Looks correct.
What's up with the formatting? And why 10 zeros lol

FinnBreu (changes requested):

src/application/imu_handler/imu_handler.c
celery6: im so blind thanks for catching that

src/application/imu_handler/imu_handler.c
celery6: auto formatting issue, will beautify later

PR #271, opened 2025-07-10
Roll program

celery6:
refactor controller.c into controller_module.c and controller.c to match matlab. This is the same as how estimator.c and estimator_module.c work.

updated design doc controller section too


src/application/controller/controller.c
```diff
@@ -44,6 +44,50 @@ static w_status_t controller_send_can(float canard_angle) {
...
+        log_text(ERROR_TIMEOUT_MS, "controller", "get_ms fail");
+        status |= W_FAILURE;
+    }
+
+    // set controller output
+    controller_output.commanded_angle = cmd;
+    controller_output.timestamp = (uint32_t)timestamp_ms;
+
+    // send command via CAN
+    if (controller_send_can(controller_output.commanded_angle) != W_SUCCESS) {
```
SAT-oO, line 71: This might be me but BARR

src/application/controller/controller.c
```diff
@@ -94,136 +139,79 @@ w_status_t controller_get_latest_output(controller_output_t *output) {
...
+            }
+
+            // get elapsed flight time
+            status |= flight_phase_get_flight_ms(&flight_time_ms);
+
+            // run controller module
+            status |= controller_module(new_state_msg, flight_time_ms, &new_cmd, &ref_signal);
+
+            // send cmd if we can
+            if (status != W_SUCCESS) {
```
SAT-oO, line 189: Barr

src/application/controller/controller.c
```diff
@@ -94,136 +139,79 @@ w_status_t controller_get_latest_output(controller_output_t *output) {
...
+                // if anything fails, send no cmd. MCB failsafes to 0 after 100ms of silence
+                log_text(ERROR_TIMEOUT_MS, "controller", "fail; send no cmd");
+            } else {
+                status |= process_new_cmd(new_cmd, 0);
```
SAT-oO, line 193: .

src/application/controller/controller.c
```diff
@@ -94,136 +139,79 @@ w_status_t controller_get_latest_output(controller_output_t *output) {
...
-                    log_text(
-                        ERROR_TIMEOUT_MS, "controller", "commanded angle failed to send via CAN"
-                    );
-                }
-                break;
-
-            default: // if not in proper state
-                vTaskDelay(pdMS_TO_TICKS(STATE_ELSE_TIMEOUT));
-                break;
+        if (controller_run_loop() != W_SUCCESS) {
```
SAT-oO, line 213: Barr

src/application/controller/controller_module.c
```diff
@@ -0,0 +1,112 @@
...
+
+w_status_t controller_module(
+    controller_input_t input, uint32_t flight_ms, double *output_angle, float *ref_signal
+) {
+    // ref_signal
+    float r = 0.0f;
+    controller_gain_t controller_gain = {0};
+
+    // validate inputs
+    if (output_angle == NULL || ref_signal == NULL) {
```
SAT-oO, line 71: Will finish the rest later

src/application/controller/controller_module.c
```diff
@@ -0,0 +1,112 @@
...
+    // compute commanded angle
+    double dot_prod = 0.0;
+    double output = 0.0;
+    for (int i = 0; i < ROLL_STATE_NUM; i++) {
+        dot_prod += gain.gain_k[i] * roll_state.roll_state_arr[i];
+    }
+
+    output = dot_prod + gain.gain_k_pre * ref_signal;
+
+    output = fmin(fmax(output, -max_commanded_angle), max_commanded_angle);
```
SAT-oO, line 58: max and min instead of max and min

SAT-oO: I thought we satisfied with output angle as floats?
dunno if this is how tis supposed to be, but reference signal is `float` and the rest are double

src/application/controller/controller_module.c
```diff
@@ -0,0 +1,116 @@
...
+    if (output_angle == NULL || ref_signal == NULL) {
+        log_text(10, "cntlmodule", "nullptrs");
+        return W_INVALID_PARAM;
+    }
+
+    // %% Reference signal
+    // % Generates reference signal for roll program
+    // % includes multiple roll angle steps. Reference r [rad].
+    if (flight_ms >= STEP_1_START_MS && flight_ms < STEP_2_START_MS) {
+        r = 0.5;
```
FinnBreu, line 86: The angles could be macros too

src/application/controller/controller_module.c
```diff
@@ -0,0 +1,116 @@
+#include "application/controller/controller_module.h"
+#include "application/logger/log.h"
+#include "common/math/math.h"
+
+// roll program steps. ms from launch time (launch is 0ms)
```
FinnBreu, line 5: Not using flight phase to define time after burnout or something?

FinnBreu (approved):
Roll program looks correct, and so does feedback law

FinnBreu (changes requested):
For flight_ms < STEP_1, this should not be a failure mode. Add on a 
`if flight_ms < STEP_1 {
   r = 0.0;
}
`

src/application/controller/controller_module.c
celery6: see controller.c

FinnBreu:
> > For flight_ms < STEP_1, this should not be a failure mode. Add on a `if flight_ms < STEP_1 { r = 0.0; } `
> 
> this is a failure because controller_module should only be called by controller after burnout. Thus if it's called before burnout, no commands should happen

The first step does not happen immediately after burnout, there is a few seconds of r=0 to zero out any roll rate accumulated during boost

SAT-oO:
> > The first step does not happen immediately after burnout, there is a few seconds of r=0 to zero out any roll rate accumulated during boost
> 
> to clarify: the STEP_... timestamps are all relative to launch=0s. However, the DTEG lockout is after boost. those 2 are independent lockouts

I think Finn's referring to the latest update he made 2 days ago lol 
they haven't been reflected in this pr yet :face

PR #285, opened 2025-07-31
v2.2.0 - v2.2.4 implementation

SAT-oO:


src/application/estimator/pad_filter.c
```diff
@@ -116,18 +117,34 @@ w_status_t pad_filter(
         return W_FAILURE; // avoid division by zero
```
celery6, line 117: delete this check if its not a potential issue anymore

celery6:
code for flight
(unless reset cmd gets fixed literally tmr...)

celery6:
(unless reset cmd gets fixed literally tmr...)
fixed! :face:
solution is to call NVIC_systemreset in the fatal err handler because apparently thats where it gets stuck after resetting.. so eventually after trying to reset enough times one of them gotta work 💀 
tested to work very reliably surprisingly

FinnBreu:
Newest version is v2.2.3

PR #287, opened 2026-02-03
Setup platformio 2026

celery6:
move firmware build to platformio!! Particular notes of interest:
- decided against using `framework = stm32cube` which would use the platformio package manager to auto-download cubemx-generated files (ex, the HAL files). Instead, we will continue to use the HAL files generated by cubemx because it turns out platformio package's HAL doesn't allow some settings that we need (like the HAL register callbacks setting). All the auto-gen files are now in `src/cubemx_autogen`
- currently you have to run `scripts/cubemx_cleanup.sh` after regenerating from cubemx. Could be automated in the future but not critical imo because it simply fails to compile if u forget so u can't accidentally forget
- new clang-format script exists to download the executable locally and run it outside of devcontainer

related changes:
- the team upgraded to clang-format-21 from 19 so i had to change the math code a bit
- adding -Wextra to compiler caught a few unused parameter warnings i removed
- we forked the xsens-mti lib to start fixing some issues it has

unit test stuff hasn't changed. It still uses devcontainer/cmake

thanks @Shiming-He for doing all the hard parts!


SAT-oO (dismissed):
workssssss 
(had some UI sync issues with the formatting button but CLI was fully functional

Shiming-He (dismissed):
Build and upload works on my Mac. Build work on cs linux server. And committed change to remove unused Apple Silicon work around that is not used.

src/common/math/math.h
```diff
@@ -9,7 +9,7 @@
 
 #include <stdbool.h>
 
-#define M_PI 3.14159265358979323846
+#define M_PI 3.14159265
```
martendo: Why did we truncate here?

PR #289, opened 2026-02-27
Flight Phase Acceleration Launch Detection

Shiming-He:
To test out using sensors for launch detection
This has now been edited so much that there is not much point in reusing this, so will be closed but good starting off point for launch detection


src/application/flight_phase/flight_phase.c
```diff
@@ -311,18 +319,78 @@ w_status_t flight_phase_update_state(flight_phase_event_t event, flight_phase_st
...
+
+	estimator_all_imus_input_t all_imu_data = imu_handler_get_data();
+
+	if (true == all_imu_data.pololu.is_dead) 
+	{
+		// TODO: Logging
+		return W_FAILURE;
+	}
+	
+	double acceleration_magnitude = math_vector3d_norm(&all_imu_data.pololu.accelerometer); // TO BE CAHNGED to ST IMU
```
celery6: confirmed with finn that acceleration magnitude is definitely the right measurement to use for this?

src/application/flight_phase/flight_phase.c
```diff
@@ -311,18 +319,78 @@ w_status_t flight_phase_update_state(flight_phase_event_t event, flight_phase_st
...
+
 /**
  * Task to execute the state machine itself. Consumes events and transitions the state
+ * Perform checks to determine if a sensor based transition is nesscary
  */
 void flight_phase_task(void *args) {
 	(void)args;
 	flight_phase_event_t event;
 	while (1) {
+		flight_phase_sensor_detection(&curr_state, &consec_num_detecion);
```
celery6: arguably this sensor_detection logic should be called inside the state machine during the relevant states only. Ex: you could call `sensor_launch_Detetcion()` during pad filter, `sensor_something_detection()` during boost or whatever later.

I fear this global `sensor_detection` style is making it too global when it doesnt need to be. For example now u have a case switch of states inside `sensor_detection` which is just duplicating the state machine switching

src/application/flight_phase/flight_phase.c
```diff
@@ -311,18 +319,78 @@ w_status_t flight_phase_update_state(flight_phase_event_t event, flight_phase_st
...
 }
 
+
+/**
+ * @brief would complete sensor-based detection of state change for flight phase
+ * @param state is a pointer to the present state
+ * @param num_consec_detection is a pointer to the number of consecutive detection made in this particular flight phase
+ * @return the status of if the function completed properly
+ */
+w_status_t flight_phase_sensor_detection(const flight_phase_state_t *state, int *num_consec_detection)
```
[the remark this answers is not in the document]
celery6: i see its un-static cuz u want to unit test it. But that might be another reason to build it into the state machine so u can test it the same way other flight phase transitions are tested rn. As opposed to calling it individually. (see other comment below about module structure)

src/application/flight_phase/flight_phase.c
```diff
@@ -81,13 +98,13 @@ w_status_t flight_phase_init(void) {
 	if (NULL == state_mailbox || NULL == event_queue || NULL == act_delay_timer ||
```
SAT-oO: https://barrgroup.com/sites/default/files/barr_c_coding_standard_2018.pdf

section 1.4a: 
> Do not rely on C’s operator precedence rules, as they may not be
obvious to those who maintain the code. To aid clarity, use
parentheses (and/or break long statements into multiple lines of
code) to ensure proper execution order within a sequence of operations.

tests/unit/flight_phase_test.cpp
```diff
@@ -346,3 +351,171 @@ TEST_F(FlightPhaseTest, RecoveryToRecovery2) {
     EXPECT_EQ(state, STATE_RECOVERY);
     EXPECT_EQ(status, W_SUCCESS);
 }
+
+TEST_F(FlightPhaseTest, PadFilterToBoostSensorDetection_DetectNominal1) {
+    // Arrange
+    flight_phase_state_t state = STATE_PAD_FILTER;
+    int consec_num_detecion = 0;
```
SAT-oO: typo

SAT-oO (changes requested):
Comments also includes things to perfect for last year's code, might as well make those changes while we're at it. A couple things:

- be careful with typos 
- comments consistency -- I suggest that you use `// TODO: things to be changed/completed/reviewed, etc`, rn I've seen `TO DO`, `TO CONSIDER`, `TODO` and so on 

very good overall

**ill insert a note here if useful someday: we don't have special fallback logic/error handle for queues failing to send events. for now its just logs and wait for next cycle

SAT-oO (changes requested):
re Christine's comments cuz not all have been addressed:

- static ur functions if they are not supposed to interface with external modules 
- more cxt-specific function names? 

(didn't check the comments in detail those are more for your own workflow. nice progress!

Shiming-He:
Currently TODO's are blocked by not having ST IMU implemented, not having a estimator get data function, and not having new Polairs data for certain constants (very low priority)

src/application/flight_phase/flight_phase.c
```diff
@@ -311,31 +346,155 @@ w_status_t flight_phase_update_state(flight_phase_event_t event, flight_phase_st
...
+w_status_t flight_phase_sensor_detection(const flight_phase_state_t *state,
+										 const estimator_all_imus_input_t *all_imu_data,
+										 const uint32_t *last_imu_timestamp,
+										 uint32_t *num_consec_detection,
+										 flight_phase_event_t *sensor_event) {
+	*sensor_event = EVENT_NULL;
+	bool threshold_detection = false;
+	flight_phase_event_t trigger_event = EVENT_NULL;
+
+	if (true ==
```
Joe-Joe-Joe-Joe: redundant

PR #290, opened 2026-03-03
move all thirdparty stuff in lib

celery6:
earlier, we wanted to add the `-w` compiler flag to the third_party libs. I used a SCons script to accomplish that.

Unfortunately, the script unintentionally prevented everything aside from third_party from being compiled (ie, src/ was not being compiled).

To fix that, 1 option was to fix the script (there's a few ways to fix it i think).

Another option is to use the `lib/` feature of platformio to actually segregate third party code from src code. It's unclear if there's any other good advantages of doing that, but it makes it easy to append specific lib-specific flags like `-w`.

And it's the canonical way of dealing with third party stuff in platformio i think


celery6:
this flopped badly. Something about linking stm32cubemx autogen stuff as a sepratae lib makes the micro crash at a weird spot always. fixed by https://github.com/waterloo-rocketry/cansw_canards/pull/292 instead

PR #299, opened 2026-03-05
Little fs

mchan5:
Include LittleFS For SD Card Logging


JasonBrave (changes requested):
1. Use git submodules to import littlefs
2. you do not need the littleFSA thing
3. delete the rocketlib submodule in root folder
4. delete the .clangd file in root, each person has their own clangd configuration, add it to gitignore

PR #302, opened 2026-03-06
Power handler header file

SuriTian:
Created header file for the power handler.

### DIFF
--- a/src/application/power_handler/power_handler.c
+++ b/src/application/power_handler/power_handler.c
@@ -9,7 +9,7 @@
 #include "rocketlib/include/common.h"
 
 /**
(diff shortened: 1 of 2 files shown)


src/application/power_handler/power_handler.c
```diff
@@ -0,0 +1,35 @@
...
+uint32_t power_handler_get_status(void) {
+    return 0;
+}
+
+/**
+ * Toggles 5V payload rail.
+ */
+w_status_t power_handler_set_payload_power(bool enabled) {
+    return W_SUCCESS;
+}
```
celery6: u also need functions to help toggle low-power mode

PR #354, opened 2026-03-15
update I2C driver to send arbitary commands and data through I2C

Shiming-He:
A method of sending arbitary data through I2C, without requirements to also send both register and data bytes in the command.

### DIFF
--- a/src/drivers/i2c/i2c.c
+++ b/src/drivers/i2c/i2c.c
@@ -196,8 +196,8 @@ w_status_t i2c_read_reg(i2c_bus_t bus, uint8_t device_addr, uint8_t reg, uint8_t
 		return W_FAILURE;
 	}
 
-	// Convert 7-bit device address to 8-bit HAL format by shifting left and setting R/W bit to 0
-	uint16_t hal_addr = (device_addr << 1) & 0xFE;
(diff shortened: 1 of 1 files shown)


src/drivers/i2c/i2c.h
```diff
@@ -81,6 +81,21 @@ w_status_t i2c_read_reg(i2c_bus_t bus, uint8_t device_addr, uint8_t reg, uint8_t
...
+ * @param[in] device_addr 7-bit device address (raw, will be shifted left internally)
+ * @param[in] data Data to write
+ * @param[in] len Number of bytes to write (max 255)
+ * @retval W_SUCCESS Write completed successfully
+ * @retval W_INVALID_PARAM Invalid parameters
+ * @retval W_FAILURE Bus not initialized
+ * @retval W_IO_TIMEOUT Transfer timed out
+ * @retval W_IO_ERROR Bus error occurred (NACK, bus error, etc)
+ */
+w_status_t i2c_send_data(i2c_bus_t bus, uint8_t device_addr, const uint8_t *data, uint8_t len);
```
celery6: nit: rename to i2c_write_data because convention

PR #382, opened 2026-03-31
add float32_t and float64_t types

celery6:
barr 
image

its fine that our previous code doesn't use this, but moving forward we will use these names

### DIFF
--- a/src/common/math/math.h
+++ b/src/common/math/math.h
@@ -13,6 +13,20 @@
 #define M_PI 3.14159265
 #endif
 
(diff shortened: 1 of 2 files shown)


src/common/math/math.h
SAT-oO: wait would our typedef cause re-def errors with the ones in here?

SAT-oO (changes requested):

SAT-oO (approved):
should we do a refactor for all the code last year? or are we just keeping this in mind for this year

PR #383, opened 2026-03-31
Movella Branch Updated for MX Configurations

AidanJCSemple:
configured MX for movella pins based on last years schematic

generated code and used the ./scripts/cubemx_cleanup.sh command as provided

formatted the code using the PlatformIO formatting tool


celery6:
moved these changes to #424 and it works!

PR #386, opened 2026-04-01
Create a customizable Rate Limiter function

### LINKED ISSUE #363: come up with better way for modules to log at a certain frequency for CAN, flash, logger
to avoid this repetitive annoying logic, which is based on last year's rushed logging code

Image

could have some helper like `rate_limiter(desired_freq, freq_which_this_is_called)` to only run every X loops of a task.

requirements:
- helper should allow a task running at `Y` Hz to run a block of code at `X` Hz, and X is easily configurable, and the block of code to run can be anything

Shiming-He:
See Issue comments, potentially not going to done or will be part of logger

### DIFF
--- a/src/common/rate_limiter/rate_limiter.c
+++ b/src/common/rate_limiter/rate_limiter.c
@@ -1,0 +1,26 @@
+
+#include <stdbool.h>
+#include <stdint.h>
+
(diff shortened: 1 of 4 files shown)


JasonBrave (changes requested):

celery6:
replaced with #401

PR #408, opened 2026-04-09
add dma for adc and continuous mode

celery6:
title
relevant reqs: 
image


src/third_party/cubemx_autogen/src/adc.c
```diff
@@ -236,7 +239,7 @@ void MX_ADC3_Init(void)
   */
   sConfig.Channel = ADC_CHANNEL_0;
   sConfig.Rank = ADC_REGULAR_RANK_1;
-  sConfig.SamplingTime = ADC3_SAMPLETIME_2CYCLES_5;
+  sConfig.SamplingTime = ADC3_SAMPLETIME_24CYCLES_5;
```
ChristineS112, line 242: whys this one set to 24.5 while the others are set to 16.5 cycles

ChristineS112 (approved):

src/third_party/cubemx_autogen/src/adc.c
celery6: adc1/adc2 are similar but adc3 is slightly different and doesnt support 16.5 so i picked 24.5 as the closest

(theyre actully diff sections in the reference manual:)

image

PR #409, opened 2026-04-09
Initialize Timer Module First and Remove Fatal Error Handler Reliance on CAN Handler Init

Shiming-He:
This is so that other modules can use timer in init.c

### DIFF
--- a/src/application/init/init.c
+++ b/src/application/init/init.c
@@ -100,6 +100,10 @@ static w_status_t init_with_retry_param(w_status_t (*init_fn)(void *), void *par
 }
 
 static void system_init_task(void *arg) {
+	// initialize timer first to make sure other modules can use it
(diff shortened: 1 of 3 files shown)


src/application/init/init.c
```diff
@@ -100,6 +100,10 @@ static w_status_t init_with_retry_param(w_status_t (*init_fn)(void *), void *par
 }
 
 static void system_init_task(void *arg) {
+	// initialize timer first to make sure other modules can use it
+	if (W_SUCCESS != timer_init()) {
+		proc_handle_fatal_error("timerinit");
+	}
```
celery6: why is this above the vtaskdelay(500)? theres no harm in putting it under

celery6 (approved):

PR #410, opened 2026-04-10
ADC Driver

ChristineS112:


src/drivers/adc/adc.h
```diff
@@ -10,7 +10,16 @@
 
 typedef enum {
 	PROCESSOR_BOARD_VOLTAGE = 0,
-	ADC_CHANNEL_COUNT
+	ADC_CHANNEL_COUNT,
```
celery6: the point of ADC_CHANNEL_COUNT is to be the sentinel value to mark the end of the enum. So it's supposed to be at the end of the list lol

ChristineS112: Lol

celery6 (approved):
yay

PR #370, opened 2026-03-24
Canardboard ioc update

celery6:
will assign people to update specific peripherals in this checklist: #361

for ppl working on this: pls create your own branch from this branch to update your peripheral, then PR to merge it into this branch

eventually we will merge this into main

expecting to merge this pr as soon as the board arrives and blinky works with this code, but other peripherals will be untested and those will be updated in detail on per-module basis later

pin info: https://docs.google.com/spreadsheets/d/1R60l7a1GPZrk1-Guafgqmg8lstQdjkIb5YvL78OFDvg/edit?gid=0#gid=0

PR #412, opened 2026-04-12
Update MX file for SDMMC

mchan5:
Updated MX File to Match new Board 
Maps SDMMC ports to corresponding pin.


celery6:
done in #370 instead

PR #413, opened 2026-04-12
Updated MX file for SDMMC

mchan5:
Updated MX file, matching SDMMC Ports with corresponding pins


celery6:
did this in #370 instead

PR #414, opened 2026-04-13
Timerioc

celery6:
10khz desired timer freq

270 Mhz base clock freq for timers

set timer period in ioc to 270,000,000 / 10,000 - 1 = 26999


Shiming-He (approved):
lgtm

PR #472, opened 2026-05-18
IIS2MDC Header

### LINKED ISSUE #481: Mag Header

michaely07:
title


src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,47 @@
...
+
+typedef struct {
+	int16_t x;
+	int16_t y;
+	int16_t z;
+} iis2mdc_raw_data_t;
+
+w_status_t iis2mdc_init(void);
+w_status_t iis2mdc_check_sanity(void);
+w_status_t iis2mdc_get_data(iis2mdc_raw_data_t *out);
```
celery6: add docstring comments for all public functions and structs like the rest of the codebase does

celery6: we need a function to get converted data too. See last year's altimu-10 driver

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,47 @@
...
+
+// Expected WHO_AM_I value for sanity checking
+#define IIS2MDC_WHO_AM_I_VAL 0x40
+
+// Register addresses
+#define IIS2MDC_REG_OFFSET_X_L 0x45 // add other axes later
+#define IIS2MDC_REG_WHO_AM_I 0x4F
+#define IIS2MDC_REG_CFG_A 0x60
+#define IIS2MDC_REG_CFG_B 0x61
+#define IIS2MDC_REG_CFG_C 0x62
```
celery6: register defines should not be in this .h file because no one else should know about it. Put this in the .c file or in a separate iis2mdc_regs.h

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,72 @@
+#ifndef IIS2MDC_H
+#define IIS2MDC_H
+
+#include "rocketlib/include/common.h"
+#include <stdint.h>
+
+// I2C slave address
+#define IIS2MDC_I2C_ADDR 0x1E
```
Shiming-He: Remember you should have to have all of these macros defined in the .c file since no one else needs this. Also these should be static const as that gives you a type check

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,72 @@
...
+ CFG_REG_A = 0x8C  COMP_TEMP_EN=1, LP=0(high-res), ODR=11 (100 Hz), MD=00 (continuous)
+ CFG_REG_B = 0x02  OFF_CANC=1 (continuous offset cancellation)
+ CFG_REG_C = 0x10  BDU=1 (block data update, DRDY pin unused, host polls status register)
+ */
+#define IIS2MDC_INIT_CFG_A 0x8C
+#define IIS2MDC_INIT_CFG_B 0x02
+#define IIS2MDC_INIT_CFG_C 0x10
+
+// conversion factor from raw register values to gauss
+const float IIS2MDC_SENSITIVITY_GAUSS_PER_LSB = 0.0015;
```
Shiming-He: Unless you’ll need to be using this in other file again this should be static const in the.c file

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,72 @@
...
+typedef struct {
+	int16_t x;
+	int16_t y;
+	int16_t z;
+} iis2mdc_raw_data_t;
+
+/**
+ * @brief Converted magnetometer data for each axis in gauss.
+ */
+typedef struct {
```
Shiming-He: You should use vector 3d for converted data as specified in common/gnc_types.h . And you should always try to use highest precision in this case so float64

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,72 @@
...
+w_status_t iis2mdc_check_sanity(void);
+
+/**
+ * @brief Retrieves magnetic field reading for all three axes.
+ * @note Polls STATUS_REG until a new sample is available then reads output registers
+ * @param[out] data Pointer to store the converted magnetic field values
+ * @param[out] raw_data Pointer to store the raw magnetometer readings
+ * @return Status of the operation, timeout if no data arrived in time.
+ */
+w_status_t iis2mdc_get_data(iis2mdc_data_t *data, iis2mdc_raw_data_t *raw_data);
```
Shiming-He: See above comment about vector 3d

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,72 @@
+#ifndef IIS2MDC_H
+#define IIS2MDC_H
+
+#include "rocketlib/include/common.h"
+#include <stdint.h>
```
Shiming-He, line 4: This should be declared in a separate section above rocketlib see https://docs.waterloorocketry.com/standards-tutorials/firmware/embedded-coding-standard.html

src/drivers/IIS2MDC/IIS2MDC.h
Shiming-He: If you want 32 bit conversion specify float32_t or if you want 64bit use float64_t. This is a Barr issue

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -0,0 +1,23 @@
+#include "drivers/IIS2MDC/IIS2MDC.h"
+#include "drivers/i2c/i2c.h"
+
+// Register addresses
+#define IIS2MDC_REG_OFFSET_X_L 0x45 // add other axes later
```
Shiming-He: Again as these are all numbers use static const

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -0,0 +1,23 @@
+#include "drivers/IIS2MDC/IIS2MDC.h"
+#include "drivers/i2c/i2c.h"
```
Shiming-He, line 2: I get it here, but in general unless otherwise stated please don’t include files you haven’t used yet. This will slow compilation and unnesscary rebuilding

Shiming-He (changes requested):
Looks good so far. After you address these changes feel free to start implementing. Just branch off of this branch

src/drivers/IIS2MDC/IIS2MDC.h
Shiming-He: Like there should be none unless there is a very good reason

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,41 @@
...
+
+#include "common/math/math.h"
+#include "rocketlib/include/common.h"
+
+/**
+ * @brief Raw magnetometer data for each axis.
+ * @note Sensitivity is 1.5 mgauss/LSB, multiply raw by 0.0015 to convert to gauss.
+ */
+typedef struct {
+	float64_t x;
```
Shiming-He: Is your raw data from the mag a float? Cause this should be whatever type that is (ie no type cast when putting the register values here). If it's an int keep it an int (make sure correct size (ie 16bit -> int16_t and 32bit -> int32_t))

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -0,0 +1,47 @@
...
+static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
+
+// conversion factor from raw register values to gauss
+static const float64_t IIS2MDC_SENSITIVITY_GAUSS_PER_LSB = 0.0015;
+
+w_status_t iis2mdc_init(void) {
+	return W_SUCCESS;
+}
+
+w_status_t iis2mdc_check_sanity(void) {
```
Shiming-He: I haven't read your datasheet but do you have a specific sanity check method?

Shiming-He (changes requested):
almost there

src/drivers/IIS2MDC/IIS2MDC.h
michaely07: wait I forgot why I even changed it to float in the first place, its int16 which I had originally

src/drivers/IIS2MDC/IIS2MDC.c
michaely07: verify the who_am_I address, verify temperature is normal via temp sensor, verify magnetometer isn't physically broken via self_test (cfg_reg_C), trigger soft_rst (cfg_reg_A)

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * @brief Performs a sanity check by verifying device identity and initialization configs.
+ * @return W_SUCCESS if the device responds with the expected values, W_FAILURE otherwise
+ */
+static w_status_t iis2mdc_check_sanity(void) {
+	return W_SUCCESS;
+}
+
+w_status_t iis2mdc_init(void) {
+	iis2mdc_check_sanity();
```
Shiming-He, line 46: Make sure in the actual implementation to have being checked for state

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include "common/math/math.h"
+#include "rocketlib/include/common.h"
+
+/**
+ * @brief Raw magnetometer data for each axis.
+ * @note Sensitivity is 1.5 mgauss/LSB, multiply raw by 0.0015 to convert to gauss.
+ */
+typedef struct {
+	int16_t x;
```
Shiming-He: How are you reading the data, cause when you process i2c messages you should be returning usigned int, so if you are doinng that this should be unsigned

src/drivers/IIS2MDC/IIS2MDC.h
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * @brief Initializes the IIS2MDC magnetometer.
+ * @note Performs a soft reset, verifies WHO_AM_I, and applies configs.
+ * @return Status of the operation
+ */
+w_status_t iis2mdc_init(void);
+
+/**
+ * @brief Retrieves magnetic field reading for all three axes.
```
Shiming-He, line 27: See Christine's comment. This needs to be a non-blocking read. So please change this to match that

Shiming-He (changes requested):
Just a few more things. Ping me once you are done.

And what's your plan for implementing this?

src/drivers/IIS2MDC/IIS2MDC.c
Shiming-He: Just a note

src/drivers/IIS2MDC/IIS2MDC.h
michaely07: I think I have to keep it int16 here because the raw data is twos complement so it can't be unsigned. Can't we just cast in the get_data function? like i2c returns uint8, cast that to uint16 then to int16

Shiming-He: I get that, but since (assuming your data is stored more than 1 register) you’ll have to combine the register data to form a uint16 (or larger) so we would just want to return this and not the typecasted version. This has less operations so this is closets to storing the raw register data that we can get

Shiming-He: Yes it’s 2’s complement, but we would still want the most raw data, so unsigned

michaely07: So we want to have it unsigned in the struct and cast to signed when getting the data? Or do you mean just have it unsigned everywhere? how would we handle negative values in that case

Shiming-He: That's an implementation thing, but look at st or ad sensor prs

Shiming-He: Basically the way it works is that you are going to be getting the raw data from combining the register data, then you will type cast to signed then to float64 (for example) to output. So bascially unsigned in the structure, but signed when we generate the actual data. 

Also you shouldn't be doing any casting or math in the get function, it should just grab the data and return it. See ST IMU

michaely07: ok I see that clears it up

src/drivers/IIS2MDC/IIS2MDC.c
michaely07: noted

src/drivers/IIS2MDC/IIS2MDC.h
michaely07: wait but doesnt the get_gyro_acc function in st imu  parse and convert raw values to physical values

Shiming-He: Yes but this data structure is used to return a raw value, so this isn’t the converted data. That’ll be outputted through vector3d as a adjusted float

Shiming-He (approved):
looks good

PR #486, opened 2026-05-29
Magnetometer driver implementation

### LINKED ISSUE #485: mag implementation
first draft: it's fine if read() blocks

ie ignoring `FIRM-LLR-SEN-003`

later will refactor to meet that req

michaely07:


michaely07:
sanity check is only necessary pre-launch right, self test is part of sanity check and it has to be blocking, so we cant run sanity checks during flight

michaely07:
should we return an error if theres no new data since the last read? currently just returning the stale data

Shiming-He:
> should we return an error if theres no new data since the last read? currently just returning the stale data

Check requirements

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,220 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+		log_text(1, "iis2mdc", "ERROR: sanity check failed");
+		return W_FAILURE;
+	}
+
 	return W_SUCCESS;
 }
 
-w_status_t iis2mdc_get_data(vector3d_t *data, iis2mdc_raw_data_t *raw_data) {
+w_status_t iis2mdc_get_data(vector3d_t *data, iis2mdc_raw_data_t *raw_data,
+							float64_t *timestamp_ms) {
```
Shiming-He: timestamp is uint32_t, please don't typecast it. It is more accurate this way

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,220 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+
+	if (W_SUCCESS != a_read(IIS2MDC_REG_OUTX_L, buf, 6)) {
+		log_text(1, "iis2mdc", "ERROR: failed to read output registers");
+		return W_FAILURE;
+	}
+
+	a_convert(buf, raw_data, data);
+
+	uint32_t now = 0;
+	(void)timer_get_ms(&now);
```
Shiming-He: you should check if it was successful

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,220 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
 
 /**
- * @brief Performs a sanity check by verifying device identity and initialization configs.
- * @return W_SUCCESS if the device responds with the expected values, W_FAILURE otherwise
+ * @brief Helper function to read one or more consecutive bytes over I2C
+ * @note IIS2MDC auto-increments the sub-address if MSB is 1, so multi-byte reads only need the
+ * start register.
+ */
+static w_status_t a_read(uint8_t reg, uint8_t *data, uint8_t len) {
+	if (len > 1) {
```
Shiming-He: I feel like we have this implemented... or at least hal does

Shiming-He (changes requested):
This is just the first part of my review so you don't get stuck waiting for me

src/drivers/IIS2MDC/IIS2MDC.c
Shiming-He: nvm, maybe this looks right but feels sus compared to other designs

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,220 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+	uint8_t status;
+
+	for (uint32_t i = 0; i < IIS2MDC_ST_TIMEOUT_MS; i++) {
+		if (W_SUCCESS != a_read(IIS2MDC_REG_STATUS, &status, 1)) {
+			return W_FAILURE;
+		}
+		if (status & IIS2MDC_STATUS_ZYXDA) {
+			return W_SUCCESS;
+		}
+		vTaskDelay(pdMS_TO_TICKS(1));
```
Shiming-He: Magic number. Not allowed in BARR

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,220 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+	data->x = (float64_t)(int16_t)raw->x * IIS2MDC_SENSITIVITY_GAUSS_PER_LSB;
+	data->y = (float64_t)(int16_t)raw->y * IIS2MDC_SENSITIVITY_GAUSS_PER_LSB;
+	data->z = (float64_t)(int16_t)raw->z * IIS2MDC_SENSITIVITY_GAUSS_PER_LSB;
+}
+
+/**
+ * @brief Polls status register until a new sample is ready (ZYXDA), waits 1ms between polls
+ * @note Bounded by IIS2MDC_ST_TIMEOUT_MS (20ms, 2 sample period at 100hz)
+ */
+static w_status_t st_wait_data_ready(void) {
```
Shiming-He: You also a have drdy gpio pin which you should use for the interrupt

src/drivers/IIS2MDC/IIS2MDC.c
michaely07: done

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+	uint8_t status;
+
+	for (uint32_t i = 0; i < IIS2MDC_ST_TIMEOUT_MS; i++) {
```
Shiming-He: This is not actually in ms, cause all of the other operation in the loop also takes time, so a loop is slightly longer than 1ms. So what you should do is either use our timer module to get the exact timestamp or use this as a loop variable. So at this point I would recommend you to just use timer module as it is more straightforward to understand

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+ * @brief Helper function to convert six output register bytes into raw counts and gauss.
+ * @note Each output register is uint8, two registers form a piece of data
+ * for an axis. This is casted to uint16 for raw data and int16 for gauss.
+ */
+static void iis2mdc_convert_sample(const uint8_t *buf, iis2mdc_raw_data_t *raw, vector3d_t *data) {
+	raw->x = (uint16_t)(((uint16_t)buf[1] << 8) | buf[0]);
+	raw->y = (uint16_t)(((uint16_t)buf[3] << 8) | buf[2]);
+	raw->z = (uint16_t)(((uint16_t)buf[5] << 8) | buf[4]);
+
+	data->x = (float64_t)(int16_t)raw->x * IIS2MDC_SENSITIVITY_GAUSS_PER_LSB;
```
Shiming-He: brackets, it is part of BARR to do all order of operation through brackets, you assume nothing

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+ * @note Bounded by IIS2MDC_ST_TIMEOUT_MS (20ms, 2 sample period at 100hz)
+ */
+static w_status_t st_wait_data_ready(void) {
+	uint8_t status;
+
+	for (uint32_t i = 0; i < IIS2MDC_ST_TIMEOUT_MS; i++) {
+		if (W_SUCCESS != iis2mdc_read_reg(IIS2MDC_REG_STATUS, &status, 1)) {
+			return W_FAILURE;
+		}
+		if (status & IIS2MDC_STATUS_ZYXDA) {
```
Shiming-He: explain why this works in a comment, and make this either equal to 0 or what ever 1 is

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+/**
+ * @brief Polls status register until a new sample is ready (ZYXDA), waits 1ms between polls
+ * @note Bounded by IIS2MDC_ST_TIMEOUT_MS (20ms, 2 sample period at 100hz)
+ */
+static w_status_t st_wait_data_ready(void) {
+	uint8_t status;
+
+	for (uint32_t i = 0; i < IIS2MDC_ST_TIMEOUT_MS; i++) {
+		if (W_SUCCESS != iis2mdc_read_reg(IIS2MDC_REG_STATUS, &status, 1)) {
+			return W_FAILURE;
```
Shiming-He, line 117: I think a timeout is better here

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+	}
+	if (W_SUCCESS != iis2mdc_read_reg(IIS2MDC_REG_OUTX_L, buf, 6)) {
+		return W_FAILURE;
+	}
+	iis2mdc_convert_sample(buf, &raw, out);
+	return W_SUCCESS;
+}
+
+/**
+ * @brief Discards the first sample, then averages 50 data samples
```
Shiming-He, line 154: add @details to explain why we doing this

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+/**
+ * @brief Runs the on-chip self-test (ST AN5080 document for procedure)
+ * @note Waits for turn-on, averages the field with self-test off, then on, and checks the
+ *       delta against the datasheet range (15-500 mgauss) Restores normal config before returning.
+ */
+static w_status_t iis2mdc_self_test(void) {
+	vector3d_t avg_off, avg_on;
+
+	// wait for stable output after power-up before sampling (specified in AN 5080)
+	vTaskDelay(pdMS_TO_TICKS(IIS2MDC_ST_POWERUP_MS));
```
Shiming-He: This pause should be in init not here, if this is required to write to register, or else lable it as such

Shiming-He (changes requested):

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+		log_text(1, "iis2mdc", "ERROR: self-test read failed");
+		return W_FAILURE;
+	}
+
+	// per-axis delta
+	float64_t dx = fabs(avg_on.x - avg_off.x);
+	float64_t dy = fabs(avg_on.y - avg_off.y);
+	float64_t dz = fabs(avg_on.z - avg_off.z);
+
+	if (dx < IIS2MDC_ST_MIN_GAUSS || dx > IIS2MDC_ST_MAX_GAUSS || dy < IIS2MDC_ST_MIN_GAUSS ||
```
Shiming-He: bracket the if statement. Order of operation BARR

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
 
 w_status_t iis2mdc_init(void) {
-	iis2mdc_check_sanity();
+	// soft reset clears config registers
+	if (W_SUCCESS != iis2mdc_write_reg(IIS2MDC_REG_CFG_A, IIS2MDC_CFG_A_SOFT_RESET)) {
+		log_text(1, "iis2mdc", "ERROR: soft reset failed");
+		return W_FAILURE;
+	}
+
+	if (W_SUCCESS != iis2mdc_write_reg(IIS2MDC_REG_CFG_A, IIS2MDC_INIT_CFG_A) ||
```
Shiming-He: BARR Order of operation see above

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+
 	return W_SUCCESS;
 }
 
-w_status_t iis2mdc_get_data(vector3d_t *data, iis2mdc_raw_data_t *raw_data) {
+w_status_t iis2mdc_get_data(vector3d_t *data, iis2mdc_raw_data_t *raw_data,
+							uint32_t *timestamp_ms) {
+	uint8_t buf[6] = {0};
+
+	if (NULL == data || NULL == raw_data || NULL == timestamp_ms) {
```
Shiming-He: BARR order of operation

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +57,224 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+
+	if (W_SUCCESS != iis2mdc_read_reg(IIS2MDC_REG_OUTX_L, buf, 6)) {
+		log_text(1, "iis2mdc", "ERROR: failed to read output registers");
+		return W_FAILURE;
+	}
+
+	iis2mdc_convert_sample(buf, raw_data, data);
+
+	uint32_t now = 0;
+	if (W_SUCCESS != timer_get_ms(&now)) {
```
Shiming-He: just use timestamp_ms since it's the right type. This is redundant and causes extra space to be allocated for the function

Shiming-He (changes requested):

src/drivers/IIS2MDC/IIS2MDC.c
michaely07: note to self dont resolve this yet

src/drivers/IIS2MDC/IIS2MDC.c
michaely07: i think the 0x80 masking thing is unique to the mag because otherwise it does not auto increment

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -233,15 +247,18 @@ static w_status_t iis2mdc_sanity_check(void) {
...
 
 w_status_t iis2mdc_init(void) {
+	// wait for stable output after power-up before any access
+	vTaskDelay(pdMS_TO_TICKS(IIS2MDC_ST_POWERUP_MS));
+
 	// soft reset clears config registers
 	if (W_SUCCESS != iis2mdc_write_reg(IIS2MDC_REG_CFG_A, IIS2MDC_CFG_A_SOFT_RESET)) {
 		log_text(1, "iis2mdc", "ERROR: soft reset failed");
 		return W_FAILURE;
 	}
```
Shiming-He, line 265: Add a pause here, after a soft reset like this, you'll need to get the powerup pause again, just to make sure that the board has set up by then

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -81,32 +82,48 @@ static w_status_t iis2mdc_write_reg(uint8_t reg, uint8_t val) {
...
+	while (now_ms - start_ms < IIS2MDC_ST_TIMEOUT_MS) {
 		if (W_SUCCESS != iis2mdc_read_reg(IIS2MDC_REG_STATUS, &status, 1)) {
 			return W_FAILURE;
 		}
-		if (status & IIS2MDC_STATUS_ZYXDA) {
+		// ZYXDA is bit 3 of STATUS_REG, it is 1 when a fresh sample is available.
+		// ORing status with IIS2MDC_STATUS_ZYXDA (1 << 3) clears every other bit, so the
+		// result is the value of just that bit's position. The
+		// negation makes it a 1 for data ready or 0 for not ready.
+		uint8_t data_ready = !(status & IIS2MDC_STATUS_ZYXDA);
```
Shiming-He: Just equate the result of the status ready bitfield with the IIS2MDC_STATUS_ZYXDA. Like you should be using the bitmap to filter to only bit 3 of STATUS_REG, but then equate to IIS2MDC_STATUS_ZYXDA to 0, and make this explicit

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -81,32 +82,48 @@ static w_status_t iis2mdc_write_reg(uint8_t reg, uint8_t val) {
...
+
+	while (now_ms - start_ms < IIS2MDC_ST_TIMEOUT_MS) {
 		if (W_SUCCESS != iis2mdc_read_reg(IIS2MDC_REG_STATUS, &status, 1)) {
 			return W_FAILURE;
 		}
-		if (status & IIS2MDC_STATUS_ZYXDA) {
+		// ZYXDA is bit 3 of STATUS_REG, it is 1 when a fresh sample is available.
+		// ORing status with IIS2MDC_STATUS_ZYXDA (1 << 3) clears every other bit, so the
+		// result is the value of just that bit's position. The
+		// negation makes it a 1 for data ready or 0 for not ready.
```
Shiming-He: make sure you code behaves as such

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +58,238 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+	uint8_t status = 0;
+	uint32_t start_ms = 0;
+
+	if (W_SUCCESS != timer_get_ms(&start_ms)) {
+		return W_FAILURE;
+	}
+
+	uint32_t now_ms = start_ms;
+
+	while (now_ms - start_ms < IIS2MDC_ST_TIMEOUT_MS) {
```
Shiming-He: brackets

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +58,238 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+		log_text(1, "iis2mdc", "ERROR: failed to restore configs after self-test");
+		return W_FAILURE;
+	}
+	if (W_SUCCESS != read_status) {
+		log_text(1, "iis2mdc", "ERROR: self-test read failed");
+		return W_FAILURE;
+	}
+
+	// per-axis delta
+	float64_t dx = fabs(avg_on.x - avg_off.x);
```
Shiming-He, line 215: make sure you include <math.h>, since this is a math.h function

Shiming-He (dismissed):
a few more things

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -1,18 +1,31 @@
...
 
-// I2C slave address
-static const uint32_t IIS2MDC_I2C_ADDR = 0x1E;
+// I2C bus and slave address
+static const i2c_bus_t IIS2MDC_BUS = I2C_BUS_4;
+static const uint8_t IIS2MDC_I2C_ADDR = 0x1E;
 
 // Register addresses
-static const uint32_t IIS2MDC_REG_OFFSET_X_L = 0x45; // add other axes later
+static const uint32_t IIS2MDC_REG_OFFSET_X_L =
```
celery6: ? idk what this question means but did u resolve this irl yet?

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +60,240 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+			return W_FAILURE;
+		}
+	}
+	return W_IO_TIMEOUT; // no new sample within the timeout
+}
+
+/**
+ * @brief Waits for a fresh sample, then reads and converts the output registers to gauss.
+ */
+static w_status_t st_read_sample(vector3d_t *out) {
```
celery6: is this Helper function only supposed to be used for self test? u should comment that so it doesn't get used in actual get_data by accident

celery6: the prefix "st" is not very clear cuz that acronym means so many things

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +60,240 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+	if (W_SUCCESS != iis2mdc_sanity_check()) {
+		log_text(1, "iis2mdc", "ERROR: sanity check failed");
+		return W_FAILURE;
+	}
+
 	return W_SUCCESS;
 }
 
-w_status_t iis2mdc_get_data(vector3d_t *data, iis2mdc_raw_data_t *raw_data) {
+w_status_t iis2mdc_get_data(vector3d_t *data, iis2mdc_raw_data_t *raw_data,
```
celery6, line 292: why doesn't this function check drdy status? Does calling read always output latest data with no corruption possible?

michaely07: with BDU enabled (cfg reg c) theres no corruption per-axis but its possible to have corruption cross-axis. However in the async implementation using the interrupt pin there shouldnt be any corruption. in this pr since we had it polling at 200hz and the mag's max ODR is 100hz i didnt want it returning failure like every other time so it just returns the newest available data even if its a little stale

src/drivers/IIS2MDC/IIS2MDC.c
michaely07: @Shiming-He do you remember what tristan said about calibration?

Shiming-He: No calibration will be done in a field at some later point. @michaely07 Can you make a very clear todo about that?

src/drivers/IIS2MDC/IIS2MDC.c
Shiming-He: ok at least add a comment about that. @celery6 are you good with leaving this like this for now?

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -1,18 +1,31 @@
+#include "FreeRTOS.h"
+#include "task.h"
+
+#include "application/logger/log.h"
 #include "drivers/IIS2MDC/IIS2MDC.h"
 #include "drivers/i2c/i2c.h"
+#include "drivers/timer/timer.h"
+
+#include <math.h>
```
Shiming-He: Nit: Please read over the Waterloo rocketry firmware guide and see where you are suppose to put the includes

src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -35,18 +60,242 @@ static const uint32_t IIS2MDC_INIT_CFG_C = 0x10;
...
+	uint8_t status = 0;
+	uint32_t start_ms = 0;
+
+	if (W_SUCCESS != timer_get_ms(&start_ms)) {
+		return W_FAILURE;
+	}
+
+	uint32_t now_ms = start_ms;
+
+	while ((now_ms - start_ms) < IIS2MDC_SELF_TEST_TIMEOUT_MS) {
```
Shiming-He, line 115: Not: add a comment on why this while loop will exit

src/drivers/IIS2MDC/IIS2MDC.c
michaely07: yup

src/drivers/IIS2MDC/IIS2MDC.c
celery6: ??? @michaely07

michaely07: uhhh shiming said to not resolve it yet cuz this is something thats gonna be implemented for async read

michaely07: ig i can resolve it now tho

src/drivers/IIS2MDC/IIS2MDC.c
celery6: image
aside from the potential cross-axis corruption (💀 ) this does meet the requirements so this is fine for now. corruption issue to be fixed in async version ig!

celery6 (approved):
works enough for now :FACE:

good progress!

PR #497, opened 2026-05-31
IIS2MDC async implementation

### LINKED ISSUE #491: Mag Implementation for Non-blockinng Get

michaely07:


src/drivers/IIS2MDC/IIS2MDC.c
```diff
@@ -22,31 +41,347 @@ static const uint32_t IIS2MDC_STATUS_ZYXDA = (1 << 3);
...
 
+/**
+ * @brief Discards the first sample, then averages 50 data samples
+ */
+static w_status_t st_collect_average(vector3d_t *avg) {
+	vector3d_t sample;
+	float64_t sum_x = 0.0, sum_y = 0.0, sum_z = 0.0;
+
+	// discard the first sample after a mode change
+	if (W_SUCCESS != st_read_sample(&sample)) {
```
celery6: https://waterloorocketry.slack.com/archives/C0A90FDAVD5/p1773767413799499

src/drivers/iis2mdc/IIS2MDC.c
Shiming-He, line 1: Why was this file moved and folder name renamed

cansw_canard_board.ioc
```diff
@@ -62,6 +62,27 @@ ADC3.OffsetSign-0\#ChannelRegularConversion=ADC3_OFFSET_SIGN_NEGATIVE
 ADC3.OffsetSign-1\#ChannelRegularConversion=ADC3_OFFSET_SIGN_NEGATIVE
 ADC3.Rank-0\#ChannelRegularConversion=1
 ADC3.Rank-1\#ChannelRegularConversion=2
+ADC3.SamplingTime-0\#ChannelRegularConversion=ADC3_SAMPLETIME_24CYCLES_5
+ADC3.SamplingTime-1\#ChannelRegularConversion=ADC3_SAMPLETIME_24CYCLES_5
```
Shiming-He: Why was this added here? You shouldn't touch anything that's not connected with your pr

cansw_canard_board.ioc
```diff
@@ -279,9 +300,9 @@ I2C2.I2C_Fall_Time=0
 I2C2.I2C_Speed_Mode=I2C_Fast_Plus
 I2C2.IPParameters=Timing,I2C_Speed_Mode,I2C_Fall_Time
 I2C2.Timing=0x00601A5A
-I2C4.I2C_Speed_Mode=I2C_Fast
+I2C4.I2C_Speed_Mode=I2C_Fast_Plus
```
Shiming-He, line 301: Please go check the datasheet when you make changes, your sensor does not support fast mode plus

cansw_canard_board.ioc
```diff
@@ -343,37 +364,38 @@ Mcu.CPN=STM32H733VGT6
...
-Mcu.IP11=FMAC
-Mcu.IP12=FREERTOS
-Mcu.IP13=I2C1
-Mcu.IP14=I2C2
-Mcu.IP15=I2C4
-Mcu.IP16=I2C5
-Mcu.IP17=MEMORYMAP
-Mcu.IP18=NVIC
-Mcu.IP19=OCTOSPI1
+Mcu.IP10=FDCAN2
```
Shiming-He, line 365: What is going on here? What are you changing?

Shiming-He, line 365: Please do not manually change the ioc file (also not with AI) EVER. This is like our master copy of all autogenerated code from cubemx. So please use cubemx

src/drivers/iis2mdc/IIS2MDC.c
Shiming-He: PLease don't rename folder just arbitrarily, other pr's rely on the places your files are at

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -4,14 +4,20 @@
...
 
 // I2C bus and slave address
 static const i2c_bus_t IIS2MDC_BUS = I2C_BUS_4;
 static const uint8_t IIS2MDC_I2C_ADDR = 0x1E;
 
+// HAL takes the 8-bit left shifted address for HAL_I2C_Mem_Read_DMA calls.
+static const uint16_t IIS2MDC_HAL_ADDR = (uint16_t)(IIS2MDC_I2C_ADDR << 1);
+
+// external definition of hi2c for i2c bus 4;
+extern I2C_HandleTypeDef hi2c4;
```
Shiming-He: PLease do this by including stm32h7xxx.h (something like this file), which will automatically include this

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -228,32 +262,79 @@ static w_status_t iis2mdc_self_test(void) {
...
+	if (hi2c != &hi2c4) {
+		return;
+	}
+
+	uint32_t timestamp_ms = 0;
+	if (W_SUCCESS != timer_get_ms(&timestamp_ms)) {
+		return;
+	}
+
+	iis2mdc_convert_sample(iis2mdc_dma_buf, &iis2mdc_cache.raw, &iis2mdc_cache.data);
```
Shiming-He: Try not to do math in interrupt functions. I know st imu does this but this is not a good way of doing this. If you can please move this to the get function instead

Shiming-He (changes requested):
A few changes

cansw_canard_board.ioc
michaely07: Im pertty sure all the ioc changes resulted from me trying to rebase, then force pushing back to the original branch, because i didnt touch anything in cubemx

src/drivers/iis2mdc/IIS2MDC.c
michaely07: i made it lowercase to match the IMU folder naming convention, lowercase folder name uppercase files names

cansw_canard_board.ioc
michaely07: Screenshot 2026-06-14 224652

src/drivers/iis2mdc/IIS2MDC.c
michaely07: moved, ISR does memcpy now

src/drivers/iis2mdc/IIS2MDC.c
michaely07: in init.c I think #include "i2c.h" does this

src/drivers/iis2mdc/IIS2MDC.c
Shiming-He: I get that, but you are causing quite a lot of issue with this (for other prs, this is not the only pr), can you just not rename it please

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -228,32 +261,78 @@ static w_status_t iis2mdc_self_test(void) {
 
 /**
  * @brief Performs a sanity check by verifying device identity and running a self test
+ * @note Blocks DMA reads during this time by setting state to CHECKING
  * @return W_SUCCESS if the device passes self test and identity verification
  */
 static w_status_t iis2mdc_sanity_check(void) {
 	uint8_t id;
```
Shiming-He: add a check to make sure you are not checking when you are already in IIS2MDC_STATE_CHECKING

Shiming-He: Clearly define this to equal something, pretty sure BARR

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -228,32 +261,78 @@ static w_status_t iis2mdc_self_test(void) {
...
+	if (iis2mdc_dma_busy) {
+		return W_FAILURE;
+	}
+	iis2mdc_dma_busy = true;
+
+	if (HAL_OK != HAL_I2C_Mem_Read_DMA(&hi2c4,
+									   IIS2MDC_HAL_ADDR,
+									   IIS2MDC_REG_OUTX_L | IIS2MDC_SUB_AUTO_INC,
+									   I2C_MEMADD_SIZE_8BIT,
+									   iis2mdc_dma_buf,
```
Shiming-He: use iis2mdc_cache.raw_buf off the bat here instead of iis2mdc_dma_buf. Since you will be copying over the data from iis2mdc_dma_buf to iis2mdc_cache.raw_buf, so it's safe doing that

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -279,35 +358,39 @@ w_status_t iis2mdc_init(void) {
 		return W_FAILURE;
 	}
 
+	// load all settings and end use of i2c driver before switching to DMA
+	vTaskDelay(1);
```
Shiming-He: First use pdMs_to_tick since vTaskDelay takes in ticks not ms. Second no magic numbers 😭

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -279,35 +358,39 @@ w_status_t iis2mdc_init(void) {
...
+
+	// register completion callback on the mag's I2C handle.
+	if (HAL_OK !=
+		HAL_I2C_RegisterCallback(&hi2c4, HAL_I2C_MEM_RX_COMPLETE_CB_ID, iis2mdc_dma_complete)) {
+		log_text(1, "iis2mdc", "ERROR: failed to register DMA complete callback");
+		return W_FAILURE;
+	}
+
+	iis2mdc_state = IIS2MDC_STATE_READY;
```
Shiming-He: This variable name is really not descriptive at all, make sure make this more discriptive. Same for the state. What does IIS2MDC_STATE_READY mean? Also how do protect you system from making sure that once you reregister the above callback that you won't use any calls that use our i2c driver, cause that will error

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -228,32 +261,78 @@ static w_status_t iis2mdc_self_test(void) {
...
+	}
+
+	memcpy(iis2mdc_cache.raw_buf, iis2mdc_dma_buf, sizeof(iis2mdc_cache.raw_buf));
+	iis2mdc_cache.timestamp_ms = timestamp_ms;
+	iis2mdc_cache.valid = true;
+
+	iis2mdc_dma_busy = false;
+}
+
 w_status_t iis2mdc_init(void) {
```
Shiming-He, line 374: protect init from being called after you have already reregistered your callback

src/drivers/iis2mdc/IIS2MDC.c
```diff
@@ -228,32 +261,78 @@ static w_status_t iis2mdc_self_test(void) {
 
 /**
  * @brief Performs a sanity check by verifying device identity and running a self test
+ * @note Blocks DMA reads during this time by setting state to CHECKING
  * @return W_SUCCESS if the device passes self test and identity verification
  */
 static w_status_t iis2mdc_sanity_check(void) {
```
Shiming-He, line 291: samething with protecting against calling this function once you have reregistered your callback

src/drivers/iis2mdc/IIS2MDC.c
Shiming-He, line 137: Both of these functions should protection to make sure neither function can be called once you switch your callback

Shiming-He, line 160: Samething about i2c protection

Shiming-He, line 1: Make sure on your file for all functions add protection to make sure your function has inited if they needed the mag to be initialized. And for those that can't be used once you reassigned your callback add that protection as well

Shiming-He: Based on testing this is causing some issues. The issue seems to be the FreeRTOS Tick count delay, so changing this to anything above 20 ms will fix this. (ie vTaskDelay means will delay up to 20 ms in this case)

Shiming-He: Same thing here as below change to above 20 ms

Shiming-He, line 225: Since you want to make sure you are reading 50 different samples not 5 samples (I assume), you will need to add a vTaskDelay of 10 ms, since you have to remember  your sensor is running at 100hz. Also please make this delay to also scale depending on which ODR we choose

src/application/init/init.c
```diff
@@ -17,7 +17,6 @@
 #include "application/init/init.h"
 #include "application/logger/log.h"
 #include "application/sensor_handler/sensor_handler.h"
-#include "drivers/IIS2MDC/IIS2MDC.h"
```
Shiming-He, line 20: Why was this deleted?

michaely07: just have everything include it as iis2mdc/IIS2MDC (lowercase folder name)

michaely07: its included like a few lines down

PR #519, opened 2026-06-13
Issue 506 motor driver unit test

maximbeauc:
## PR Overview
Added unit tests for ak45_driver for Issue 506 Motor Driver Unit Test #518 
- `Added hal_fdcan_mock.h`, `hal_fdcan_mock.c`, and stub `fdcan.h` for FDCAN HAL mocking
- Added `ak45_driver_reset` and `ak45_set_tx_errors` to `ak45_driver.c`/`.h` for help with unit testing
- Added unit tests for all public functions in `ak45_driver.h`: `ak45_driver_init`, `ak45_send_position_cmd`, `ak45_send_disable_cmd`, `ak45_get_latest_feedback`, and `ak45_get_tx_errors`

### Notes
First PR in Rocketry, so please give feedback even if not specific to these changes.
For context, this PR is from a Fork since I'm not a collaborator.

### DIFF
--- a/src/drivers/ak45_driver/ak45_driver.c
+++ b/src/drivers/ak45_driver/ak45_driver.c
@@ -297,17 +297,3 @@ void HAL_FDCAN_RxFifo1Callback(FDCAN_HandleTypeDef *hfdcan, uint32_t RxFifo1ITs)
 		ak45_fdcan_rx_callback(hfdcan, RxFifo1ITs);
 	}
 }
-
-/* for unit testing use */
-void ak45_driver_reset(void) {
-	g_ak45_hfdcan = NULL;
-	g_feedback_queue = NULL;
-	g_tx_errors = 0;
-	is_init = false;
-	received_can_msg = false;
-}
-
-/* for unit testing use */
-void ak45_set_tx_errors(uint32_t count) {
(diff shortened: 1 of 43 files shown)


src/drivers/ak45_driver/ak45_driver.c
Shiming-He, line 1: For now, for our unit tests we are not going to be adding unit test functions in our src files. So try your best to be able to edit things without this, but if you can't thats fine

maximbeauc: I just couldn't find a way to get around resetting the static variables between tests. Is there a way to do that with FFF or other things we have? I can get rid of the function that sets the number of errors pretty easily though.

Shiming-He: I'm not sure, but if there isn't one then just don't test it

maximbeauc: Is it okay if I move the functions into a separate helper file which is in the unit tests folder? That way I don't modify the preexisting .c file. I had it the other way around because it made more sense to me, but I can see why you wouldn't want unit test functions in the src files.

Shiming-He (approved):
lgtm

PR #524, opened 2026-06-14
Use canard firmware error msg for health check 

### LINKED ISSUE #517: Improve health checks error reporting format
Must improve module error reporting format over CAN and text logging to be more human readable. Currently sends general board status message with msg string "module_id:error_bitfield". Should instead use new CANARD_FIRMWARE_ERROR message over CAN.

ChristineS112:


src/application/health_checks/health_checks.c
Shiming-He: Can this be moved to the top of the file with the rest of the constants

Shiming-He (commented):
Rest lgtm

src/application/health_checks/health_checks.c
```diff
@@ -159,17 +159,23 @@ static w_status_t process_module_status(health_status_t status) {
 				 status.severity,
 				 status.error_bitfield);
 
-		uint8_t data[6] = {0};
 		can_msg_t msg = {0};
 		// build error msg in the form "module id:error bitfield:severity"
-		snprintf_((char *)data, sizeof(data), "%d:%lu", status.module_id, status.error_bitfield);
-		build_debug_raw_msg(PRIO_HIGH, xTaskGetTickCount(), data, &msg);
+		build_canard_firmware_error_msg(PRIO_HIGH,
```
JasonBrave:
```suggestion
		build_canard_firmware_error_msg(PRIO_MEDIUM,
```

src/application/health_checks/health_checks.c
JasonBrave, line 281: You need to set all bit field bits that applicable to you

ChristineS112: wdym

Shiming-He: Like right now you are just changing a single bit for the failure any or all of the modules? is that the design requirement?

ChristineS112: yes

ChristineS112: we send general board status msg to indicate any overall failure and firmware error msg for which specific module failed

src/application/health_checks/health_checks.c
```diff
@@ -159,38 +174,29 @@ static w_status_t process_module_status(health_status_t status) {
 				 status.severity,
 				 status.error_bitfield);
 
-		uint8_t data[6] = {0};
 		can_msg_t msg = {0};
 		// build error msg in the form "module id:error bitfield:severity"
-		snprintf_((char *)data, sizeof(data), "%d:%lu", status.module_id, status.error_bitfield);
-		build_debug_raw_msg(PRIO_HIGH, xTaskGetTickCount(), data, &msg);
+		build_canard_firmware_error_msg(PRIO_MEDIUM,
+										xTaskGetTickCount(),
```
JasonBrave: xTaskGetTickCount returns milliseconds right?

Shiming-He: No that returns tick counts, which in our design is 1 tick = 1 ms, but this is not standard, so if we decide to go with 1 tick = 0.5 ms, then this is broken

Shiming-He: please switch to get time from timer module

src/application/health_checks/health_checks.c
Shiming-He: samething here please switch to use timer module

src/application/health_checks/health_checks.c
JasonBrave: the timestamp must be in milliseconds

Shiming-He: yup, so have commented to switch to time from our timer module

celery6: for future reference, our entire codebase is supposed to use timer module for time, nothing at all should use tickcount
this is written somewhere... I think

